PR #6, opened 2017-02-10
Attempt to integrate razor view engine

nicolocodev:
I hope this closes the #1 

I googled about razor engine and F# and found some interesting projects like [FSRazor](https://github.com/dahlbyk/FSRazor), [Shaver](https://github.com/Dzoukr/Shaver) and [RazorEngine](https://github.com/Antaris/RazorEngine). But none of those is .net core 😢 Then, i found a [nice C# implementation](https://github.com/Antaris/RazorEngine) based on [SO this answer](http://stackoverflow.com/a/38568835/1229323) so i decided to use this.

By now you can configure where your views (`.cshtml`s) are placed: `let view = configureRazor "views"` or use a default config with `let view = defaultRazor ()`, then you call your `view` function like `view "Person.cshtml" model` 

There are some pitfalls, like you must use `namespaces` on your models other way razor can't find the `@model` on `.cshtml`.


dustinmoris:
Hi,

I wanted to let you know that I saw your pull request and I already started to look into it.

You are right, it seems to be difficult to find a Razor library for .NET Core at the moment. RazorLight looks cool, but I saw a [potential issue](https://github.com/toddams/RazorLight/issues/31) with it which I raised with the maintainer now.

Apart from that I like what you have done. There's a few things I would suggest to improve (e.g. have `IRazorLightEngine` as a dependency so we don't have to instantiate it every time) but I will give more detailed feedback when I have more clarity regarding the async/await issue I raised.

Many thanks for helping out here and taking the initiative to resolve #1. I appreciate it a lot and will do my best to get this polished with you before merging.

Just FYI I am travelling tomorrow and will be back on Sunday late night, so might not be able to respond or work on this for the next 2 days, but will get back to it as soon as possible. Many thanks again!

src/AspNetCore.Lambda/HttpHandlers.fs
```diff
@@ -275,4 +277,15 @@ let htmlFile (relativeFilePath : string) =
...
+let razor (engine : RazorLightEngine) (view : string) (model : obj) =
+    let view = engine.Parse(view, model)
+    setHttpHeader "Content-Type" "text/html"
+    >>= setBodyAsString view
+
+let configureRazor (viewsFolder : string)  = 
+    let views = combinePaths (Directory.GetCurrentDirectory()) viewsFolder
+    razor (EngineFactory.CreatePhysical(views))
+
+let defaultRazor () = configureRazor ""
```
dustinmoris: I was thinking that instead of having to create the `RazorLightEngine` every time you could merge all 3 functions to one like this perhaps:

```
let razorView (viewName : string) (model : obj) =
    fun (ctx : HttpHandlerContext) ->
        let engine = ctx.Services.GetService<IRazorLightEngine>()
        let view = engine.Parse(viewName, model)
        setHttpHeader "Content-Type" "text/html"
        >>= setBodyAsString view
        <| ctx
```

Here the HttpHandler tries to resolve an `IRazorLightEngine` from the service provider.

You could create an extension method like this:

```
type IServiceCollection with
    member this.AddRazorEngine (viewsFolderPath : string) =
        this.AddSingleton<IRazorLightEngine>(EngineFactory.CreatePhysical(viewsFolderPath));
```

... which then can be used like this:

```
        services.AddRazorEngine("C:\\AspNetCore.Lambda.SampleApp\\views") |> ignore
```

I think this would be more incline with existing ASP.NET Core patterns and it would allow users to configure the engine only once as a singleton or provide a custom implementation.

What do you think?

tests/AspNetCore.Lambda.SampleApp/project.json
```diff
@@ -6,9 +6,12 @@
     "compilerName": "fsc",
     "compile": {
       "includeFiles": [
+        "Person.fs",        
         "Program.fs"
       ]
-    }
+    },
+    "embed": [ "**/*.cshtml" ],
```
dustinmoris: When using the physical RazorLight engine then the views don't need to be embedded and this line can be removed.

tests/AspNetCore.Lambda.Tests/project.json
```diff
@@ -6,9 +6,12 @@
     "compile": {
       "includeFiles": [
         "FormatExpressionTests.fs",
+        "Models.fs",
         "HttpHandlerTests.fs"
       ]
-    }
+    },
+    "preserveCompilationContext": true,
+    "embed": [ "Person.cshtml" ]
```
dustinmoris: Here as well.

tests/AspNetCore.Lambda.SampleApp/Person.fs
```diff
@@ -0,0 +1,7 @@
+namespace AspNetCore.Lambda.SampleApp
+
+//Model
+[<CLIMutable>]
+type Person = {
+    Name : string
+}
```
dustinmoris: This is only formatting, but to be consistent I'd prefer to stick with this style please:

```
type Person = 
    {
        Name : string
    }
```

src/AspNetCore.Lambda/project.json
```diff
@@ -35,7 +35,8 @@
         "Microsoft.AspNetCore.Hosting": "1.1.0",
         "Microsoft.AspNetCore.Diagnostics": "1.1.0",
         "Newtonsoft.Json": "9.0.1",
-        "DotLiquid": "2.0.64"
+        "DotLiquid": "2.0.64",
+        "RazorLight": "1.0.0-rc1",
```
dustinmoris: Haven't tested it, but could imagine that the trailing `,` might cause a json formatting error?

src/AspNetCore.Lambda/HttpHandlers.fs
nicolocodev: i'm not creating the engine on each call (i think so), those functions `configure` and `default` returns a partial applied function with the engine, so, in the app you do `let view = razor ()` and now the engine is created once and you can use it wherever `view "foo" model` But, however, hey! I like your approach, uses the built in DI and looks more "aspnet friendly" 😸

dustinmoris: Oh yeah looking back at the code you're right, it was instantiated only once.

nicolocodev:
hi, sorry for delay, was a busy week. I've updated the implementation following all of your comments on my last commit. Let me know is there is any other issue on this! ❤️

dustinmoris:
Hi, thanks! I am currently out but will have a look later tonight!

src/AspNetCore.Lambda/Common.fs
```diff
@@ -19,4 +19,7 @@ let serializeXml x =
     let serializer = new XmlSerializer(x.GetType())
     use stream = new MemoryStream()
     serializer.Serialize(stream, x)
-    stream.ToArray()
\ No newline at end of file
+    stream.ToArray()
+
+let combinePaths (path1 : string) (path2 : string) =
+    Path.GetFullPath(Path.Combine(path1, path2))
```
dustinmoris: This method is probably not needed anymore.

src/AspNetCore.Lambda/HttpHandlers.fs
```diff
@@ -275,4 +277,12 @@ let htmlFile (relativeFilePath : string) =
...
\ No newline at end of file
+        }
+
+let razorView (viewName : string) (model : obj) =
+    fun (ctx : HttpHandlerContext) ->
+        let engine = ctx.Services.GetService<IRazorLightEngine>()
+        let view = engine.Parse(viewName, model)
+        setHttpHeader "Content-Type" "text/html"
+        >>= setBodyAsString view
+        <| ctx
```
dustinmoris: Let's add some XML documentation to the handler. Example:

```
/// Parses and compiles a Razor view with the associated model and then writes its content to the body of the response.
/// It also sets the HTTP header Content-Type to text/html.

nicolocodev:
Hi again, man, thanks for your guidance and patience on this PR! I've added the comments and updated the readme as you suggested. Let me know if there is any other issue on this!

dustinmoris:
Hey no worries. There's been many changes happening since the last merge again, so there are more conflicts again which prohibit me from merging your PR. This weekend I will take everything you've done and update a branch with it and then probably merge it back into dev/master from there if that is okay with you?

nicolocodev:
sure!

nicolocodev:
Hi! i think i get done the merge with `develop` branch! 🎉 those was a lot of changes, yay `fsproj`! I had some troubles, though:
- I've changed the `<TargetFramework>` to `netstandard1.6` 'cause `1.3` was generating the error `The target "CreateManifestResourceNames"  does not exist`
- I've added the line <EnableDefaultCompileItems>false</EnableDefaultCompileItems> 'cause i was getting the error `Duplicate 'Compile' items were included.`

Even though my tests, both of them, main tests and sample app's tests, are broken... can't find the View! i've added a `todo` line on the sampleapp in order to get some advice... buuut the sample app is working yay!

dustinmoris:
Thanks for the effort!!!

Issue #7, opened 2017-02-12
Nesting route combinators


slogsdon:
As a newcomer to AspNetCore.Lambda, I've been working on exploring the library. During this process, I attempted to do something like this:

```fsharp
let app =
  choose [
    route "/ping" >>= text "pong"
    route "/" >>= text "index"
    route "/api"
      >>= choose [
        route "" >>= setStatusCode 204 >>= text "No Content"
        route "/users" >>= text "users"
      ]
    route "/docs" >>= text "docs"
  ]
```

expecting a request to `/api` would return a `204 No Content` response and a request to `/api/users` would return `"users"`. To my surprise, I received a `404 Not Found` for both. After doing some digging, I discovered that `route` tests if the given path matches the request path, so with `route "/users"` above, it compares a request to `/api/users` with the given `/users`, fails, and returns `None`.

I wasn't able to find anything in the existing code base to handle this, so I added the following as a helper module in my application to achieve the expected behavior:

```fsharp
let optionally pred value =
  if pred then Some value else None

let getCurrentRoot (ctx : HttpContext) : string =
  if ctx.Items.ContainsKey "rootPath"
  then ctx.Items.Item "rootPath" |> string
  else ""

let rootPath (part : string) =
  fun (ctx : HttpHandlerContext) ->
    let root = getCurrentRoot ctx.HttpContext
    ctx.HttpContext.Items.Item "rootPath" <- root + part
    ctx
    |> Some
    |> async.Return

let subPath (part : string) =
  fun (ctx : HttpHandlerContext) ->
    let fullPath = (getCurrentRoot ctx.HttpContext) + part
    ctx
    |> optionally (fullPath = ctx.HttpContext.Request.Path.ToString())
    |> async.Return
```

allowing the following to work as expected:

```fsharp
let app =
  choose [
    route "/ping" >>= text "pong"
    route "/" >>= text "index"
    rootPath "/api"
      >>= choose [
        subPath "" >>= setStatusCode 204 >>= text "No Content"
        subPath "/users" >>= text "users"
      ]
    route "/docs" >>= text "docs"
  ]
```

Does something like this exist today? If not, would it be useful to include this functionality as a part of AspNetCore.Lambda?

slang25:
I would love to see something like this. I would prefer it to be just one method, but because you are flowing the path through `HttpContext` you have to have this rigid ordering of these functions. Not sure if there's another way to achieve this?

dustinmoris:
[...]

PR #12, opened 2017-02-15
Add EndsWith variant of route filters

ErikSchierboom:


dustinmoris:
Hi, thanks for submitting this pull request. Could you perhaps explain the use case for this handler a little bit? I am asking because there is already a similar discussion in issue #7 and that might even make the use case for `routeStartsWith` redundant. Would love to better understand where this might come handy so we make the right decisions.

ErikSchierboom:
@dustinmoris Ah, I didn't see that discussion. It's probably far better to go that route, sorry about that!

dustinmoris:
No worries and thanks for the effort though! Feel free to join the other discussion if you have more ideas/suggestions :)

PR #17, opened 2017-02-19
Use functions rather than classes in sample app

slang25:
Just a preference thing, a little for F# idiomatic IMO


dustinmoris:
Sweet, didn't know you can `.Configure(..)`!

Issue #20, opened 2017-02-19
Add a dotnet new template


dustinmoris:
See [here](https://blogs.msdn.microsoft.com/dotnet/2017/02/07/announcing-net-core-tools-updates-in-vs-2017-rc/#user-content-dotnet-new) for more information.

xdaDaveShaw:
There's a wiki article [here](https://github.com/dotnet/templating/wiki/%22Runnable-Project%22-Templates) describing the template layout

PR #27, opened 2017-03-12
Add a dotnet new template #20

dsincl12:
A simple dotnet new Giraffe template


dsincl12:
You install it with: ``dotnet new -i template``

template/_GiraffeTemplate/Program.fs
```diff
@@ -0,0 +1,68 @@
...
+    let host =
+        WebHostBuilder()
+            .UseKestrel()
+            .UseContentRoot(Directory.GetCurrentDirectory())
+            .Configure(Action<IApplicationBuilder> configureApp)
+            .ConfigureServices(Action<IServiceCollection> configureServices)
+            .ConfigureLogging(Action<ILoggerFactory> configureLogging)
+            .Build()
+    host.Run()
+    0
```
dustinmoris: Let's make this a bit more compact in the template:

```
let main argv =                                
    WebHostBuilder()
        .UseKestrel()
        .UseContentRoot(Directory.GetCurrentDirectory())
        .ConfigureServices(Action<IServiceCollection> configureServices)
        .ConfigureLogging(Action<ILoggerFactory> configureLogging)
        .Build()
        .Run()
    0
```

template/_GiraffeTemplate/Program.fs
```diff
@@ -0,0 +1,68 @@
...
+
+// ---------------------------------
+// Web app
+// ---------------------------------
+
+let webApp = 
+    choose [
+        GET >=>
+            choose [
+                route  "/"           >=> razorView "Hello.cshtml" { Text = "Hello, Giraffe world!" }
```
dustinmoris: I think the whitespace between the route and the razorView handler must be from a copy paste. It's not an issue, but since this will be the template let's keep it clean :)

template/_GiraffeTemplate/Program.fs
```diff
@@ -0,0 +1,68 @@
...
+open Models
+
+// ---------------------------------
+// Error handler
+// ---------------------------------
+
+let errorHandler (ex : Exception) (ctx : HttpHandlerContext) =
+    let loggerFactory = ctx.Services.GetService<ILoggerFactory>()
+    let logger = loggerFactory.CreateLogger "ErrorHandlerLogger"
+    logger.LogError(EventId(0), ex, "An unhandled exception has occurred while executing the request")
```
dustinmoris: The first two lines can be removed now, because ctx has a default `Logger` now: `ctx.Logger.LogError...`

dustinmoris: I will update the sample app with the same changes as well

dustinmoris:
Nice work! I had only a few minor comments otherwise this looks good to me! One more thing though, what do you think of renaming the default cshtml file to `Index.cshtml` since it is set for the `/` route in the template?

dsincl12:
Sure it makes sense to name it Index.cshtml. I would also love to see some more examples like a REST API for a simple CRUD app with EF, file uploads and token authentication with Identity. The project really has potential so a website and more documentation would be awesome. I really love the Rocket site https://rocket.rs. They have an overview, guide and API part which makes it really simple to get started. Maybe some milestones are in order? Thank you for your awesome work.

dustinmoris:
Added #28

dustinmoris:
Fixes #20

dustinmoris:
The rocket website looks really nice indeed. Very well laid out and the documentations explains the framework really well. Regarding giraffe I didn't think that its own website with a lot of documentation was needed, because it is essentially only a thin framework on top of ASP.NET Core, which already has a lot of documentation online and therefore I hoped that a simple API reference like in the README + a few good sample projects might be enough, but I am open for other suggestions. I'll have a think about it :)

Issue #32, opened 2017-04-05
Creating the html in a functional way


nojaf:
I did a quick experiment with using  https://github.com/SuaveIO/suave/blob/master/src/Experimental/Html.fs to construct the html.

```fsharp
let webApp = 
    choose [
        GET >=>
            choose [
                route "/" >=> razorView "Index.cshtml" { Text = "Hello, Giraffe world!" } ;
                route "/test" >=> setBodyAsString (Suave.Html.samplePage |> Suave.Html.htmlToString)
            ]

        setStatusCode 404 >=> text "Not Found" ]
```

Including the single Html.fs file in your project is enough to make this work. No need to reference the full `Suave.Experimental` nuget package.

Should Giraffe include something similar by default as alternative to Razor/dotliquid?

dustinmoris:
Hi, that is an interesting alternative indeed. I wonder how useful this is in its current form beyond a hello-world example. Looks like Suave was working on it but deemed it experimental. Is this something you need personally for your projects? 

When I look at the currently supported html tags and attributes then there's still lots of gaps. If we want to include this into Giraffe, then there needs to be an easy way for people to extend this without much difficulty, because I doubt we will ever be able to satisfy everyone's needs. My understanding is that I can easily create more tags outside the library and it will work, but am I correct that a tag can only have one attribute with this engine at the moment?

dustinmoris:
It's kind of cool.. and yeah it looks that it would be easily extendable. If you want to prepare a pull request for it then go for it!

Could you please make sure that we don't violate any copyright on Suave though and probably should let them know that we would like to borrow their code or at least heavily attribute it to the original author and give them credit!

nojaf:
Ok, I'll try to create a pul request and check with Suave for copyright.
I want to use this in a hobby project.
Another reason I'm found of this way of working is that it is similar to how Elm creates html.

```elm
view : Model -> Html Msg
view model =
  div []
    [ input [ type_ "text", placeholder "Name", onInput Name ] []
    , input [ type_ "password", placeholder "Password", onInput Password ] []
    , input [ type_ "password", placeholder "Re-enter Password", onInput PasswordAgain ] []
    , viewValidation model
    ]
```

dustinmoris:
Hey, I wanted to let you know I've made the following changes:

- Named it `Giraffe.HtmlEngine`
[...]

PR #35, opened 2017-04-07
Initial attempt for Html Nodes

nojaf:
As discussed in https://github.com/dustinmoris/Giraffe/issues/32, the port of Suave experimental Html.
I've added a sample, unit test and updated the README.
Let me know if anything is missing.
Cheers!

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,7 @@ The old NuGet package has been unlisted and will not receive any updates any mor
     - [dotLiquidHtmlView](#dotliquidhtmlview)
     - [razorView](#razorview)
     - [razorHtmlView](#razorhtmlview)
+    - [htmlNode](#htmlNode)
 - [Custom HttpHandlers](#custom-httphandlers)
 - [Installation](#installation)
 - [Sample applications](#sample-applications)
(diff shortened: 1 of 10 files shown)


dustinmoris:
Thanks for the PR. It looks good to me. I might maybe clean up a few things before merging it into master and creating a new release (e.g. the `samplePage` in the `Giraffe.Html` module is probably not required) and perhaps spend a bit more time thinking about the http handler name, but otherwise this looks pretty good as is!

I was thinking if perhaps this:
```
route  "/htmlNode"   >=> htmlNode (person { Name = "Html Node" })
```

would be maybe nicer to read like this:

```
route  "/person"   >=> person { Name = "Html Node" } |> renderHtml
```

so perhaps rename `htmlNode` to `renderHtml` or something similar. Just want to spend some time thinking about good naming before I create a new release to avoid later breaking changes. What do you think?

I'll probably create a new release with this and some more changes tomorrow or latest on Monday.

dustinmoris:
Fixes #32

nojaf:
Good idea, no need to rush anything. I agree that `htmlNode` isn't the best name.

Issue #31, opened 2017-04-04
Changes to Razor files don't seem to come through when the app is running.


nojaf:
When editing the Index.cshtml while the app is running, no changes seem to have any effect. The html output stays the same.

## Steps to reproduce

- Create a new app with `dotnet new giraffe`
- Edit the `Index.cshtml` by adding some styling

```razor
@model yellow.Models.Message

<!DOCTYPE html>
<html>
<head>
    <title>Giraffe</title>
    <link href="https://fonts.googleapis.com/css?family=Lato:300" rel="stylesheet">
    <style>
        body {
            background-color:#f1c40f;
            color:#2c3e50;
            font-family: 'Lato', sans-serif;
        }

        h3{
            font-size: 4rem;
            font-weight: 300;
            text-align: center;
            /*border:1px solid blue;*/
        }
    </style>
</head>
<body>
    <div>
        <h3>@Model.Text</h3>
    </div>
</body>
</html>
```
- run the app
- uncomment the css
- reload browser

The title doesn't seem to have a border.
I'm not sure if this is a Razor Engine issue or caching inside Giraffe.

dustinmoris:
Hi, thanks for reporting this. That is an interesting issue, not sure how to fix it though because after some initial investigation it seems like it is a level of caching in the razor engine. I'll have a closer look over the next few days and see what options I have to provide a better experience.

nojaf:
I don't think it's the razor. I noticed similar problems with the htmlEngine as well.
If the razorView has a different model, the page does get a proper refresh.

```fsharp
let getText () =
    let guid = Guid.NewGuid().ToString()
    sprintf "Hello, Giraffe world! %s" guid 

let webApp = 
    choose [
        GET >=>
            choose [
                route "/" >=> razorView "Index.cshtml" { Text = getText () } ;
            ]
```

Is this an F# thing that if you call functions without arguments or without different argument values the result of the function is cached? Because I'm guessing in strict way calling the function with the same arguments should return the same value.

Are we missing a `warbler` function somewhere?

dustinmoris:
Hi, I think it is both. In the case of the html engine you need a `warbler`, because functions in F# are eagerly evaluated and in the case of razor a `warbler` will help to evaluate the function every time, but the razor engine will still compile a view only once and changing the `cshtml` while the application is running will not have an effect.

Try this example:

```fsharp
let warbler f a = f a a

let getName() = Guid.NewGuid().ToString()

let webApp = 
[...]

PR #37, opened 2017-04-13
Added warbler and updated documentation

nojaf:
Added some documentation and an example where `warbler` is relevant.
Let me know what you think.


src/Giraffe/HttpHandlers.fs
```diff
@@ -447,4 +447,7 @@ let negotiateWith (rules : IDictionary<string, obj -> HttpHandler>) (responseObj
             <| ctx
 
 let negotiate (responseObj : obj) =
-    negotiateWith defaultNegotioationRules responseObj
\ No newline at end of file
+    negotiateWith defaultNegotioationRules responseObj
+
+let warbler f a =
```
dustinmoris: I think the warbler can be made `inline`.

nojaf: Indeed, looks like [suave](https://github.com/SuaveIO/suave/blob/b8d098f0142db23806d44a7e5555606ccaaa89dd/src/Suave/WebPart.fs#L59) did this as well. I'll change it.

README.md
```diff
@@ -784,6 +784,41 @@ let app =
...
+```
+
+Functions in F# are eagerly evaluated and in the case of razor/htmlEngine a warbler will help to evaluate the function every time.
+
+```fsharp
+// ('a -> 'a -> 'b) -> 'a -> 'b
+let warbler f a =
+    f a a
+```
+
```
dustinmoris: This is a good example. I was thinking perhaps we should move the section `Gotchas` beneath the list of default http handlers or perhaps move it under `warbler` in the http handlers, because it applies really to all handlers and not just the html or razor views. The same rules apply for `text`, `json`, `xml`, `negotiate`, etc.
What do you think?

nojaf: Good point, it's indeed bigger than html/razor. Perhaps it's time to start a wiki as the readme will soon be getting out of hand don't you think?

tests/Giraffe.Tests/HttpHandlerTests.fs
```diff
@@ -1187,4 +1187,11 @@ let ``Get "/flex" without an Accept header returns a JSON object`` () =
...
-        Assert.Equal("application/json", ctx.HttpContext.Response |> getContentType)
\ No newline at end of file
+        Assert.Equal("application/json", ctx.HttpContext.Response |> getContentType)
+
+[<Fact>]
+let ``Warbler function should execute inner function`` () =
+    let inner (a: int) = a.ToString()
+    let warbled = warbler (fun (a:int) -> inner)
+    let result = warbled 42
+    Assert.Equal("42", result)
```
dustinmoris: I think if you would remove the `warbler` the test would still pass because the function would always return 42. A good test would be to have `let inner = Guid.NewGuid().ToString()` and then call it twice without warbler and confirm that the result is the same and then call it twice with the warbler and assert that the two results are not equal.

README.md
dustinmoris: Yes that is true... 
I was thinking about the big README and so far I actually thought its not so bad. As long as the index is always updated it has actually quite some advantages, because people can easily find stuff in a singe document by simple ctrl+f search, easier edit the document with find/replace and so on. I see it a bit like a big word document, which is essentially also just one file.

I will think about a separate wiki though and perhaps move things out from the README...

dustinmoris: Another advantage is that if the API changes I can ask contributors to update the doc/README part of the PR and so I can always guarantee that the documentation is up to date. With a separate wiki I think there is a bigger risk of things getting out of date over time.

tests/Giraffe.Tests/HttpHandlerTests.fs
nojaf: Something like 
```fsharp
[<Fact>]
let ``Warbler function should execute inner function twice`` () =
    let inner = Guid.NewGuid().ToString()
    let warbled = warbler (fun _ -> inner)

    let r1 = inner
    let r2 = inner
    let r3 = warbled
    let r4 = warbled

    Assert.Equal(r1, r2)
    Assert.NotEqual(r3, r4)
```

Doesn't compile since inner is a `string` and it should be `a' -> b'`. Adding unit to inner will execute twice so I'm not sure how to proceed.

README.md
nojaf: Agreed, it's still early days so let's keep things simple. I'll move the `Gotchas` section.

tests/Giraffe.Tests/HttpHandlerTests.fs
dustinmoris: I was thinking something like this:

```
[<Fact>]
let ``Warbler function should execute inner function twice`` () =
    let ctx, hctx = initNewContext()
    let inline warbler f a = f a a
    let inner() = Guid.NewGuid().ToString()
    let app = 
        GET >=> choose [ 
            route "/foo"  >=> text (inner())
            route "/foo2" >=> warbler (fun _ -> text (inner())) ]

    ctx.Request.Method.ReturnsForAnyArgs "GET" |> ignore
    ctx.Request.Path.ReturnsForAnyArgs (PathString("/foo")) |> ignore
    ctx.Response.Body <- new MemoryStream()

    let result1 = 
        hctx
        |> app
        |> Async.RunSynchronously
        |> (fun res -> getBody res.Value)

    ctx.Response.Body <- new MemoryStream()

    let result2 = 
        hctx
        |> app
        |> Async.RunSynchronously
        |> (fun res -> getBody res.Value)

    Assert.Equal(result1, result2)

    ctx.Request.Path.ReturnsForAnyArgs (PathString("/foo2")) |> ignore
    ctx.Response.Body <- new MemoryStream()

    let result3 = 
        hctx
        |> app
        |> Async.RunSynchronously
        |> (fun res -> getBody res.Value)

    ctx.Response.Body <- new MemoryStream()

    let result4 = 
        hctx
        |> app
        |> Async.RunSynchronously
        |> (fun res -> getBody res.Value)

    Assert.False(result3.Equals result4)
```

dustinmoris:
Hey so overall this looks good and really thankful for your help, not just this PR but already throughout the investigation in the issue you have logged. I saw you already started making some more changes based on the comments, so let me know when you are all done and then I'll merge it.

tests/Giraffe.Tests/HttpHandlerTests.fs
nojaf: Great test, that does indeed cover what we are planning on using the `warbler` for.

nojaf:
Ok, I think I've got everything in place. Glad we were able to resolve #31. 
It's nice working on this project, happy to help.

dustinmoris:
Awesome, many thanks!!

Issue #30, opened 2017-04-01
Serving arbitrary files with correct headers - like Suave Files.browseHome


mtnrbq:
Is this capability planned, or is there a function for this that I've missed?

dustinmoris:
Hi, I didn't add such a function because [ASP.NET Core can already serve static files](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/static-files). Just like you configure `UseGiraffe(webApp)` in the `Startup` class you can configure `UseStaticFiles()`.

mtnrbq:
Doh! - of course - thanks.

PR #42, opened 2017-04-19
Add handler to serve any file's contents (not just html)

claruspeter:
I added this functionality to my own sandbox project, and then thought that it would be generally useful.

It allows the dev to server the contents of any static file to the response with a given content header.  The htmlFile handler already exists, but doesn't allow me to serve images/styles/scripts/pdfs etc that are useful in a standard content-rich website.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,7 @@ The old NuGet package has been unlisted and will not receive any updates any mor
     - [negotiate](#negotiate)
     - [negotiateWith](#negotiatewith)
     - [htmlFile](#htmlfile)
+    - [fileContent] (#fileContent)
     - [dotLiquid](#dotliquid)
     - [dotLiquidTemplate](#dotliquidtemplate)
     - [dotLiquidHtmlView](#dotliquidhtmlview)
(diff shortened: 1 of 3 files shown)


dustinmoris:
Hi Peter, this looks a lot like static file handling that already exists in ASP.NET Core out of the box. See also #30 where this has been mentioned. Is that the use case where you needed this?

claruspeter:
Thanks Dustin.  Yes, you're right the static file handler does this - I should RTFM a bit harder ;)  

BTW, I did have to give it a bigger hint than usual:     
````fsharp
app.UseStaticFiles( 
   StaticFileOptions(
      FileProvider = new PhysicalFileProvider(Directory.GetCurrentDirectory())
   )
) |> ignore
````
... without setting current directory it failed to find any files.  I dont' recall having to do this in other projects.  Not sure if this is an issue, yet?

In any case feel free to reject and close this PR.

Cheers, Peter.

**Comment by dustinmoris** [comment]
Hi, your snippet is one way to configure it, the other way would be by setting the WebRoot path like this:

```fsharp
[<EntryPoint>]
let main argv =     
    let contentRoot = Directory.GetCurrentDirectory()
    let webRoot = Path.Combine(contentRoot, "webroot")                       
    WebHostBuilder()
        .UseKestrel()
        .UseContentRoot(contentRoot)
        .UseWebRoot(webRoot)
        // other configurations
        .Build()
        .Run()
    0
```

`ContentRoot` sets the root folder of your application for any content files that should not be publicly accessible, but your application will use (e.g. razor views). `WebRoot` is the directory which will be browsable when you configure the static file middleware.

After setting `UseWebRoot(...)` you can simply configure the static file middleware with:

```fsharp
let configureApp (app : IApplicationBuilder) = 
    app.UseGiraffeErrorHandler(errorHandler)
    app.UseStaticFiles() |> ignore
    app.UseGiraffe(webApp)
```

Let me know if that works for you :)

Issue #46, opened 2017-04-27
Installing Giraffe on the full .NET framework


nojaf:
I'm not entirely sure this is a problem with Giraffe but I'm unable to install Giraffe on a full framework project using .NET 4.6.1. Which according to [this table](https://docs.microsoft.com/en-us/dotnet/articles/core/tutorials/libraries) also should support Net standard 1.6.

```
Install-Package : Could not install package 'Giraffe 0.1.0-alpha015'. You are trying to install this package into a project that targets '.NETFramework,Version=v4.6.1', but the package does not contain any assembl
y references or content files that are compatible with that framework. For more information, contact the package author.
At line:1 char:16
+ Install-Package <<<<  Giraffe -Pre 
    + CategoryInfo          : NotSpecified: (:) [Install-Package], Exception
    + FullyQualifiedErrorId : NuGetCmdletUnhandledException,NuGet.PackageManagement.PowerShellCmdlets.InstallPackageCommand
```
It is also unclear to me what the **with tooling 2.0** bit is in the table.

dustinmoris:
Hi, yeah that is a strange one. Did you try to add it via NuGet in VS 2017? Based on the table I would expect that with VS 2017 and via the default add nuget package dialog it should work, assuming that the "default" Microsoft way with their latest tools would work. If that is true, then you can work backwards to figure out what the problem is in your case.

nojaf:
Yes using VS 2017 with the dialog and Package Manager Console. No luck.

nojaf:
I'm able to install all the dependencies of Giraffe so something is missing in Giraffe itself.
I did notice the the [csproj of MvcCore](https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc/Microsoft.AspNetCore.Mvc.csproj#L7) mentions `net46` where [Giraffe](https://github.com/dustinmoris/Giraffe/blob/develop/src/Giraffe/Giraffe.fsproj#L10) doesn't.

Could that be the missing link?

dustinmoris:
Hmmm yes I think this could be it. Do you want to quickly try it out locally and if it works I can quickly add it and publish a new package.

nojaf:
Ok, I'll give it a shot.

dustinmoris:
Released with the latest version. Thanks!

nojaf:
Thanks for the release, I can confirm that the `0.1.0-alpha016` package works with net461.

PR #47, opened 2017-04-27
Add support for full .NET framework

nojaf:
Changing the target frameworks did the trick for #46 .
Not sure what will happen on the CI server.

Also creating an old school `fsproj ` console app and getting AspNetCore in general to work on it is a pain.
Some issues I came across:
- Libuv was not copied, see https://github.com/aspnet/KestrelHttpServer/issues/1652
- System.Numerics.Vectors gave a strange assemblyRedirect problem, I had to change my fsproj to
```xml
    <Reference Include="System.Numerics.Vectors">
   <HintPath>..\packages\System.Numerics.Vectors.4.3.0\lib\net46\System.Numerics.Vectors.dll</HintPath>
      <Private>True</Private>
    </Reference>
```
See https://github.com/dotnet/corefx/issues/3103
And a redirect in app.config

```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <startup>
    <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.6.1" />
  </startup>
  <runtime>
    <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
      <dependentAssembly>
        <assemblyIdentity name="System.Numerics.Vectors" publicKeyToken="b03f5f7f11d50a3a" culture="neutral" />
        <bindingRedirect oldVersion="0.0.0.0-4.99.0.0" newVersion="4.1.2.0" />
      </dependentAssembly>
    </assemblyBinding>
  </runtime>
</configuration>
```

In case anyone every tries to do the same.

### DIFF
--- a/src/Giraffe/Giraffe.fsproj
+++ b/src/Giraffe/Giraffe.fsproj
@@ -1,4 +1,5 @@
 ﻿<Project Sdk="FSharp.NET.Sdk;Microsoft.NET.Sdk">
+
   <PropertyGroup>
     <AssemblyName>Giraffe</AssemblyName>
     <VersionPrefix>0.1.0-alpha015</VersionPrefix>
@@ -46,7 +47,5 @@
     <Compile Include="ModelBinding.fs" />
     <Compile Include="Middleware.fs" />
   </ItemGroup>
-
-
 
 </Project>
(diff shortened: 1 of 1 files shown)


dustinmoris:
Hi, thank you, sorry for the late merge - I was away until today!

nojaf:
No problem, thanks for merging.

Issue #61, opened 2017-06-20
Separate nuget packages for View Engines


JonCanning:

JonCanning:
I'm happy to have a go at this

dustinmoris:
Hey, thanks for this ticket and the pull request. Why would you prefer the view engines to have as separate NuGet packages?

I had the same thought in the beginning and I thought about the pros/cons of both options and came to the conclusion that going with a single NuGet package is probably the preferred choice, but I am open for discussion.

## Option 1 - Single NuGet package

### Pros:
- Getting started is super easy, you only have to add one NuGet package and got everything you need
- Versioning is simple with one NuGet package, because there's only one package and only one version
- Managing code and future additions are simple. There's no ambiguity if a new package should be created or not, because the core library contains everything. For example should the `HtmlEngine` also be split out? If someone wants to add support for other serialization protocols, should they be added to the core library or a new package? Things get messy and often there is no clear line.
- No dependency updating hell. If the core library gets changed then all other libraries have to be updated to reference the latest, which is a pain.

### Cons:
- The core library will most likely contain more functionality than a single web application will need*.

*) This would be a problem with more packages as well though. An app which doesn't use razor views or XML might not add the `Giraffe.Razor` NuGet package, but will still have the `xml` http handler available.

## Option 2 - Multiple NuGet packages

### Pros:
- Core package will be smaller in size with less default functionality (the second can be seen as a con as well though)

### Cons:
- Users must search/know what packages to add for whatever functionality they need
- Versioning becomes less intuitive. If there is a fix in the core library and the core library increases the version, should the other packages match that version, even though they might not have a change or shall the versions diverge?
- If the versions diverge, would it not make more sense to split the separate NuGet packages into their own respective repositories, so that their versions, tags and releases can be properly recorded in GitHub? Separate repositories would add a lot of development overhead though.
- It is never going to be clear when to create a new repo/NuGet package. What is the criteria? How big a new feature is, if it deserves a new namespace/module shall it be moved into a new package altogether or when it introduces other new dependencies shall it be split out?
[...]

PR #62, opened 2017-06-21
Separate projects for View Engines

JonCanning:
#61 

Please review

### DIFF
--- a/Giraffe.sln
+++ b/Giraffe.sln
@@ -1,0 +1,86 @@
+
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio 15
+VisualStudioVersion = 15.0.26124.0
+MinimumVisualStudioVersion = 15.0.26124.0
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{1DD50551-DD32-4006-8B14-ACA57036C7AF}"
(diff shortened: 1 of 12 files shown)


nojaf:
I believe the template will require some modifications as well.
And I would add some text in the [razor](https://github.com/dustinmoris/Giraffe/blob/develop/README.md#razorview) and [dotliquid](https://github.com/dustinmoris/Giraffe/blob/develop/README.md#dotliquid) sections that these functions require an additional nuget package.

dustinmoris:
Hey, thanks for this PR, I put down some thoughts in #61 which I would like to discuss before making any more progress on this PR.

dustinmoris:
Ok, I have merged your work and done a few minor changes to it as part of the merge:
- Extended README with these changes
- Changed module names to `Giraffe.Razor.xxx` and `Giraffe.DotLiquid.xxx`
- Removed the `readFileAsString` function from the `Giraffe.DotLiquid` package so that it doesn't create a conflict or override the default function from Giraffe
- Removed the `setBody` function from the `Giraffe.Razor` package, because it would override the original `setBody` handler from Giraffe and it has a different signature
- Included the two new projects in the build script so that AppVeyor will produce and publish the new NuGet packages

Let me know if you see any issues with these changes and thanks for your work!

PR #63, opened 2017-06-29
TryGetQueryStringValue

JonCanning:
Just an extension I've found useful in my code

### DIFF
--- a/src/Giraffe/HttpContextExtensions.fs
+++ b/src/Giraffe/HttpContextExtensions.fs
@@ -41,8 +41,8 @@ type HttpContext with
         
     member this.TryGetQueryStringValue key =
         match this.Request.Query.TryGetValue key with
-        | true, values -> Ok values.[0]
-        | _ -> sprintf "query string value '%s' is missing." key |> Error
+        | true, values -> Some values.[0]
+        | _ -> None
 
     /// ---------------------------
     /// Model binding
--- a/src/Giraffe/HttpContextExtensions.fs
+++ b/src/Giraffe/HttpContextExtensions.fs
@@ -38,6 +38,11 @@ type HttpContext with
         match this.Request.Headers.TryGetValue(key, strValue) with
         | true  -> Ok (strValue.Value.ToString())
         | false -> Error (sprintf "HTTP request header '%s' is missing." key)
+        
+    member this.TryGetQueryStringValue key =
+        match this.Request.Query.TryGetValue key with
+        | true, values -> Ok values.[0]
+        | _ -> sprintf "query string value '%s' is missing." key |> Error
 
     /// ---------------------------
     /// Model binding


dustinmoris:
Nice, I think that is a useful extension, also it could be used inside `BindQueryString<'T>()`. There's only one thing which I would like to change and that would be to return an `Option<string>` instead of `Result<string, string>`. The reason being is because there is already a similar method `TryGetRequestHeader` which returns an option, so I would like to either stay consistent, where a user can expect a `TryGet...` method to return an `Option<'a>` or rename one of the methods. Perhaps a `GetRequestHeader` could return a `Result<string, string` as the name suggests that you definitely expect a value back and if you don't then it's sort of an error, while a `TryGetRequestHeader` is more defensive and the name suggests that it could return no result.

What do you think?

JonCanning:
Ha, it was an `Option` but I changed it to a `Result` when I saw `GetRequestHeader` so was trying to make it fit in. I should have looked a little bit further....

gerardtoconnor:
Just to be difficult, I suggest `struct (bool*'T)` ... consistant with TryGet `out` double matches, short lived, no GC pressure and no casting/additional obj ref lookups. I use it in parse lookups as I can assign a premade `struct(false,Unchecked.defaultof<'T>)` value for fails ... Just to be difficult as I said ;)

dustinmoris:
@gerardtoconnor What you say is interesting. TBH I don't know enough about it atm and to be consistent with what we already have I'm happy with the change by @JonCanning.

Quickly reading [through this](https://fslang.uservoice.com/forums/245727-f-language/suggestions/6193110-add-structoption-t) I think I need to research this topic a bit more myself, because there are some interesting comments there:
> None is represented as null in compiled code due to [<CompilationRepresentation(CompilationRepresentationFlags.UseNullAsTrueValue)>] applied on Option<'T>. So, only Some<'T> values are heap allocated.

JonCanning:
@gerardtoconnor How would you implement this?

gerardtoconnor:
@dustinmoris It's not a massive impact, I've just been playing around with the same thing trying to get some extra juice out of routing algo, please don't take my suggestion too seriously!

The complier does erase out None to null which is great and tbh, for ref types we could just do ugly null checks but by using struct check value and reference, we maintain `TryGet` pattern & safety while we remove 1 single object lookup & object creation on Some case. For struct value (which is mainly what I was using in parsers) the return type should be embedded/wrapped into struct tuple so no further look up needed and no intermediate object created.

I think Option is fine in most cases, especially for public api, I was just joking round ... but for internal performance critical result passing through pipelines, we should be using structs for small return values that are short lived and should not be sitting around on heap (this can be done where needed internlly) if makes sense.

@JonCanning (trivial example), could be extended with generics
```f#
let rtrn (v) = struct (true ,v) 
let failure  = struct (false, 0)
let f age = if age > 50 then failure else rtrn age

match f 32 with
| struct(true, v) ->
| struct(false,_) ->
```

JonCanning:
Thanks, that's a great tip

gerardtoconnor:
> Thanks, that's a great tip

np, its most useful when you are returning multiple struct value items from functions in which case performs alot better then a deep ref tuple option.
`struct(bool*int*int*float*float)` to communitcate (is result?, index1,index2,value1,value2) or some combo of values

PR #64, opened 2017-07-03
Fixed incorrect code in dotnet template

GraanJonlo:
Hi, spotted a small mistake in the dotnet template code


dustinmoris:
oh yes, thanks for spotting and fixing!

dustinmoris:
It's deployed now, many thanks for detecting it so quickly and fixing it!

PR #65, opened 2017-07-03
Implement routeMap

JonCanning:
Something else I've been using in my own project. It converts a route with parameters

`/test/{foo}/{bar}`

into a `Map<string, string>` so you can then read the parameters in your route handler. I've also implemented the `?` operator for ease

`fun map -> sprintf "%s %s" map?foo map?bar |> text`

### DIFF
--- a/src/Giraffe/Giraffe.fsproj
+++ b/src/Giraffe/Giraffe.fsproj
@@ -45,5 +45,5 @@
     <Compile Include="HttpHandlers.fs" />
     <Compile Include="Middleware.fs" />
   </ItemGroup>
-  
+
 </Project>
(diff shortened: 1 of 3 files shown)


JonCanning:
I don't know how much value this adds over routef, it's just something I used in an OWIN project some time ago so thought it might be of use

JonCanning:
Maybe I could extend this a little further to become a routeBind<'T>. Would that be of more use?

dustinmoris:
Hi, this looks useful, I will review it a bit later today and put some thought to your question regarding `routeMap` vs `routeBind<'T>`. It might even be that both is useful.. I have to think about it a bit. Just a quick other question, it seems that every file you edit you also automatically format the contents like removing redundant whitespace, which is really nice, because I miss that feature in VS Code right now. Are you using a different IDE or is there a feature in Code which I didn't see yet?

JonCanning:
`"files.trimTrailingWhitespace": true`

dustinmoris:
Looks good, thanks for this addition!

PR #72, opened 2017-08-01
fixed typo dotnet-new template

marukami:


dustinmoris:
Hi @marukami, thanks for this PR. Quick question, did you run into an issue by executing the command without the quotes? I tested it without quotes and it has always worked without any problems to me, and other examples on the internet didn't use the quotes either, but I might be missing something?

gulshan:
I think the quote is required in earlier dotnet cli versions.

marukami:
I'm running zsh on macOS and I'm on dotnet core CLI 1.0.4. Running without quotes fails on zsh 
```bash
➜  ~ dotnet new -i giraffe-template::*
zsh: no matches found: giraffe-template::*
```
I tried today with bash on macOS and Linux without quotes and it worked. It seems that zsh needs quotes.

I also saw that the [dotnet/templating](https://github.com/dotnet/templating/wiki/Available-templates-for-dotnet-new) wiki was using quotes. I would assume some other shells like zsh don't work without quotes.

dustinmoris:
Thanks, learned something new again! :)

Issue #44, opened 2017-04-22
Deploying to Azure App Service


kirse:
I had to slightly modify the base Giraffe template in order to successfully deploy it to an Azure App Service:

1. (Optional) In the Azure Portal, first clear out all the "Default Documents" in the "Application Settings" for the targeted App Service.  May also need to delete them from wwwroot.
1. Added a PackageReference to the fsproj for Microsoft.AspNetCore.Server.IISIntegration (1.1.*)
1. Added UseIISIntegration() to the WebHostBuilder init
1. Published the app with a web.config similar to the one below (alongside app DLL; app DLL name will vary)

`<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <system.webServer>
    <handlers>
      <add name="aspNetCore" path="*" verb="*" modules="AspNetCoreModule" resourceType="Unspecified" />
    </handlers>
    <aspNetCore processPath="dotnet" arguments=".\giraffe.dll" stdoutLogEnabled="false" stdoutLogFile=".\logs\stdout" />
  </system.webServer>
</configuration>`

With those changes, anyone should be able to follow [this ASP.NET Core Azure tutorial](https://www.visualstudio.com/en-us/docs/build/apps/aspnet/aspnetcore-to-azure) and deploy the `giraffe-template` Nuget pkg to their Azure App Service.

Caveats: Not really an ASP.NET Core expert yet, so I'll let you decide how to manage that web.config situation.  It gets generated into "PubTmp" if you publish a C# ASP.NET Core app from VS2017, so I'm not sure at what stage it'd be created in a vscode workflow (`dotnet publish` doesn't appear to generate it).  Also, it's possible to directly host w/ Kestrel on Azure, but general rec is to use IIS/Nginx as a reverse proxy.

dustinmoris:
Thanks for this great summary! I was thinking to create a setting for the giraffe template to create different applications. I'll turn this into an Azure web app template!

alfonsogarciacaro:
Thanks a lot @kirse! I couldn't make the deployment work until I found this issue. In case it helps others, I also [wrote a guide](https://gist.github.com/alfonsogarciacaro/578ed3cadf091121715e461c8902ac9f) for deploying Giraffe apps to Azure. It'll be nice to have an option in the template to add `web.config` indeed.

dustinmoris:
Thank you @alfonsogarciacaro! I will add the web.config to the templates, good call!

[...]

PR #78, opened 2017-08-10
Added IIS Packages for Azure Deployment

odytrice:
Considering Azure/IIS Deployment Scenarios. #44 I updated the template to support those scenarios out of the box


dustinmoris:
Great stuff, thank you!

PR #85, opened 2017-08-19
fix warbler after api change

ian-moore:
`warbler` didn't execute every time as it was missing the second argument for the new `HttpHandler` signature. Tested this out in the sample app and it works as expected now.

What I'm not sure of is why the exist unit test still passed and what would be a good way to re-write so it fails when expected.

### DIFF
--- a/samples/SampleApp/SampleApp/Program.fs
+++ b/samples/SampleApp/SampleApp/Program.fs
@@ -124,7 +124,7 @@ let webApp =
                 route  "/fileupload" >=> razorHtmlView "FileUpload" ""
                 route  "/person"     >=> (personView { Name = "Html Node" } |> renderHtml)
                 route  "/once"       >=> (time() |> text)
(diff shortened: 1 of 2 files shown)


ian-moore:
Damn, you beat me to it :) Nevermind!

dustinmoris:
Lol, yeah but many thanks for doing it! Also check my change then you can see what was the reason for the test not failing. Basically in a real world app we apply the `next` continuation only once in the middleware and therefore all the handlers got pre-compiled. In the test we applied the next continuation alongside the ctx every time and therefore didn't run into this issue.

dustinmoris:
I changed the warbler in such a way that it is a non breaking change for existing Giraffe apps, as in you don't need to write `fun _ _ -> someHandler` but can leave it as `fun _ -> someHandler`

ian-moore:
I'm not sure which use case is more common, but this is a breaking change for anyone (like me) who was using the context argument - just need to deconstruct the tuple now instead.

Issue #88, opened 2017-08-24
htmlFile Handler fails with "cannot access a closed file"


TOTBWF:
dotnet --info:
```
.NET Command Line Tools (1.0.4)

Product Information:
 Version:            1.0.4
 Commit SHA-1 hash:  af1e6684fd

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  10.12
 OS Platform: Darwin
 RID:         osx.10.12-x64
 Base Path:   /usr/local/share/dotnet/sdk/1.0.4
```

I think the issue may be in `readFileAsString` disposing of the stream or reader before the task completes.

dustinmoris:
Thanks for the fix. I just released it with the newest NuGet package: Giraffe 0.1.0-beta-003!

PR #90, opened 2017-08-24
Fix readFileAsString disposing of resources before completion.

TOTBWF:
readFileAsString was disposing of the stream and the reader before the task completed. This PR fixes that by wrapping the whole thing in a Task CE. (See issue #88)


dustinmoris:
Thanks!

PR #92, opened 2017-08-26
Fix Razor is not affected by the _ViewStart file

lukethenuke:


dustinmoris:
Fantastic, thanks for this nice fix and adding tests for it!

PR #97, opened 2017-09-24
[RFC] Add ability to customize JSON serializer settings

maxdeviant:
Currently the `json` HTTP handler will always use the default JSON serializer settings.

It is possible to change them by directly modifying `JsonConvert.DefaultSettings`, but since this will change the settings globally there are some cases where this is undesirable. It would be much better to be able to customize the serializer settings on a case-by-case basis.

From an API standpoint, do we want to make this the default for `json`? Or should there be something like a `customJson` that someone can use if they want to opt-in to this functionality?

### DIFF
--- a/src/Giraffe/Common.fs
+++ b/src/Giraffe/Common.fs
@@ -27,7 +27,11 @@ let readFileAsString (filePath : string) =
 /// Serializers
 /// ---------------------------
 
-let inline serializeJson x = JsonConvert.SerializeObject x
+let inline serializeJson (serializer : JsonSerializer) x =
+    use stream = new MemoryStream()
+    use writer = new StreamWriter(stream)
+    serializer.Serialize(writer, x)
+    stream.ToString()
 
 let inline deserializeJson<'T> str = JsonConvert.DeserializeObject<'T> str
 
--- a/src/Giraffe/HttpHandlers.fs
+++ b/src/Giraffe/HttpHandlers.fs
@@ -293,7 +293,7 @@ let text (str : string) : HttpHandler =
     setHttpHeader "Content-Type" "text/plain"
     >=> setBodyAsString str
 
-let private makeJsonHandler serializer (dataObj : obj) : HttpHandler =
+let makeJsonHandler serializer (dataObj : obj) : HttpHandler =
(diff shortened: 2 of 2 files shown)


maxdeviant:
@dustinmoris Would like to get your input before I go too far down this path :smile:

dustinmoris:
Hi, first of all sorry that I am getting back a bit delayed. I am abroad and was in a place without WiFi for the last two days :(. Secondly I think what you say makes sense and there should be a way to customise the JSON serialisation.

I like your changes, but I am wondering if the `customJson` is even needed or if it wouldn't be better to expose the `makeJsonHandler` publicly so that a user could initilize a custom serializer only once and then create a handler with it themselves which they could name whichever way they like.

Not sure, what do you think?

maxdeviant:
@dustinmoris No worries!

Exposing `makeJsonHandler` is actually how I would prefer to do it, but I also wasn't sure if that aligned with your vision of the public API for Giraffe. But this is exactly why I asked :smile:

I'll ping you once I've cleaned some things up and gotten the tests passing.

dustinmoris:
Hi, how are you getting on with this? If there's anything I can do let me know, otherwise I'll wait for you to clean up the stuff as you suggested :)

maxdeviant:
@dustinmoris Been pretty busy with work lately, so I haven't had any time to look at it. But I'll let you know if you have any questions on it.

dustinmoris:
@maxdeviant sure no rush... I really just wanted to make sure you were not waiting for me on anything :) Thanks for your help!

dustinmoris:
Hi, just FYI I think this is an important change which I would like to get completed before I consider Giraffe RC ready. Please let me know if you think you still want to work on this PR otherwise I can take over from here and finish it :)

maxdeviant:
@dustinmoris Sure thing! I'll see if I can get it merge-ready in the next couple of days.

maxdeviant:
@dustinmoris Sorry for the delay on this. I haven't had the time to focus on it, so it might be better if you take over.

dustinmoris:
Done by @KerryRJ in #146

PR #100, opened 2017-10-01
Configuring CORS.

torhovland:
Demonstrates how to enable the CORS mechanism in ASP.NET Core.

### DIFF
--- a/samples/IdentityApp/IdentityApp/IdentityApp.fsproj
+++ b/samples/IdentityApp/IdentityApp/IdentityApp.fsproj
@@ -12,6 +12,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Cors" Version="2.0.*" />
     <PackageReference Include="Microsoft.AspNetCore.Hosting" Version="2.0.*" />
     <PackageReference Include="Microsoft.AspNetCore.Diagnostics" Version="2.0.*" />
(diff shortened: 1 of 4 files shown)


nojaf:
@torhovland are you not missing the CORS package in the fsproj?

torhovland:
You're right. I've now added the ref to both the sample and the template fsproj files.

dustinmoris:
Thanks for the work! This looks good, I am not sure if we want to completely open CORS by default in the template though. Maybe it's just good enough to have it in the sample app where users can look up how to configure it when they need it? What do you think?

dustinmoris:
I'll merge it into develop anyway now and we can make that change before I merge everything back to master. Thanks for helping out, much appreciated!

torhovland:
Hi! I haven't opened CORS by default, only for http://localhost:8080, meant as an example. I somewhat agree that this may be better suited in a sample than in the template, but on the other hand, there is the argument about having the code at hand when you do actually need to configure CORS. It comes down to whether you want the template to be concise or comprehensive. As it's only 4 lines of code, I think it's fine, but I don't have any strong feelings about it.

hakonrossebo:
@torhovland - According to the solution in this Giraffe CORS issue - should we add a route similar to this in the sample to make this work from the browser? route  "/options" >=> setHttpHeader "Allow" "GET, OPTIONS, POST"

https://stackoverflow.com/questions/46725807/issues-with-http-post-despite-enabling-cors-in-both-elm-and-giraffe/46735081#46735081

PR #101, opened 2017-10-04
A sample application showing how to use a JSON Web Token (JWT) for au…

torhovland:
…thentication in ASP.NET Core with Giraffe.


dustinmoris:
Nice, thank you so much!

PR #103, opened 2017-10-06
Configuration sample

torhovland:


dustinmoris:
Awesome, thanks so much!

nojaf:
I'm a bit skeptical about this sample. It doesn't really strike me as a how-to in Giraffe.
Is this not more a missing docs thing on the aspnetcore docs site?
What is the advantage of this instead of getting the configuration from the context?

```fsharp
let myApp (next:HttpFunc) (context:HttpContext) =
    task {
        let configuration = context.GetService<IConfigurationRoot>()
        let msg = configuration.["HelloMessage"]
        return! (text msg) next context
    }
```

torhovland:
The purpose of the PR was to demonstrate a way to access configuration variables in Giraffe, as both the docs and samples were missing that. I agree my solution didn't look like idiomatic Giraffe code, and I was actually hoping somebody would eventually improve on it. Your suggestion to resolve IConfiguration within a handler is nicer, and I have submitted a new PR that simply demonstrates this in a new route within the SampleApp.

Issue #56, opened 2017-06-07
Router & Route Performance


gerardtoconnor:
@TheAngryByrd @dustinmoris @imetallica
This is a move of the topic over from Task CE issue. Just to summarise

1. Current Giraffe router uses a process of elimination on a list of path substrings/patterns, iteratively testing everyone till match met which is elegant, simple but not performant.
2. For performance, routes can be compiled into a tree structure that minimises the number of failed tested paths.
3. The best tree structures are Tries (character/edge nodes) or Binary/AVL Tree (chunk/token tested higher/lower) 
4. I have initially tried trie as it only checks path char once while Bin/AVL checks same start path, multiple chars, multiple times, so if the trie node traversal can be sped up close to string checking, trie will be faster. 
5. I have built first draft of Trie Router and it reads path, char by char checking for node path and firing off Http Handler functions at appropriate endpoints.
6. The initial draft (with exception of one route bug) has same perf for short routes, 20% faster on long, 50% on long parse routes, leading to blended 9% improvement on total.
7. Custom made crawl/stream parsers are an important part of performance boost over standard regex.
8. The initial draft uses ref obj nodes with dictionary edges, a crude and bulky way to initially test function, now that concept works, I will migrate nodes to structs with compressed sparse array edges in an effort to improve performance as the stack is faster than heap etc.
9. You can note from perf table below, there is only a slight degradation of performance on Trie over longer routes while old router understandably is noticeably slower.
10. I think the ultimate goal is that Giraffe router should be a lot faster than MVC such that Giraffe is the most compelling web platform when performance & composability are taken into account.

With all this said I will just re-iterate how Trie router Handler works vs standard route HttpHandler:
```f#
// (Node->Node) list -> HttpHandler
let routeTrie (routeFns:(Node->Node) list) = ...
// string HttpHandler -> (Node -> Node)
let routeT (path:string) (fn:HttpHandler) = fun (Node) -> Node
let (==>) (a: HttpHandler->Node->Node) (b:Httphandler) = a b 

let trieApi : HttpHandler =
    routeTrie [
        routeT "/" ==> text "Hello world, from Giraffe!"
        routeT "/test" ==> text "Giraffe test working"
        routeT "/about" ==> text "Giraffe about page!"
        routeTf "/data/%s/weather" (fun v -> sprintf "json (weatherForecasts (%s))" v |> text)
        routeTf "/value/%s" text 
[...]

PR #104, opened 2017-10-07
Token router

gerardtoconnor:
As discussed in #56, initial hybrid radix tree router submission. There is some refactoring to do and one test fails but need to understand test as I'm not sure why url encoding is being tested.

Failed test is:
>  Failed   Giraffe.TokenRouterTests.GET "/foo/blah blah/bar" returns "blah blah"
> Error Message:
> Expected: blah%20blah
> Actual:   blah blah

Readme updates are basic so feel free to tweak & refine.


gerardtoconnor:
@dustinmoris  let me know if you have any questions on the implementation. I am open to tweaking it to ease usage/adoption. 

Let me know on the `Blah Blah` test if you would like me to update to expect space instead of URL encoded %20 so that the tests pass, or would like to leave failed as is.

This is not my final planned PR on performance Router but thought I would start integrating it asap so namespace usage of `BasicRouter` vs `TokenRouter` would be less breaking later on.

dustinmoris:
Hi @gerardtoconnor,

I have to admit that I still haven't gotten my head fully around this PR, but generally I think it looks good. It needs a lot of tidy up though (seen a lot of comments indicating that you still need to finish a few things) and I'd prefer to keep the basic routes in the current namespace and then let a user override them by opening up the TokenRouter namespace.

Since the token router implementation is an opt-in feature which would get enabled through namespace opening I see little risk with this PR.

Few things which I'd like you to ask as part of this PR:
- Keep basic routes in original namespace (up for debate if you can convince me otherwise :D)
- Add to the README a bit of a description what the TokenRouter is (as in how it works under the covers) and when a user should opt in for it instead of the normal routes (basically make it more clear for less experienced users which to pick).
- Perhaps make it mandatory/optional for a user to configure a 404 handler for the TokenRouter, as I can imagine that many people will want to be able to customise the 404 response with performance routing as well
- Make helper functions such as `rtrn`, `between`, etc. private or perhaps move into a sub module of `TokenRouter` so that these functions do not accidentally override the same named functions by a user who uses TokenRouter.
- Restore all the metadata of the Giraffe.fsproj which has been removed in this PR
- Fix the last broken unit test :)

Overall great work and really appreciate your many contributions!

gerardtoconnor:
On comments, given the complexity of execution, is it ok to leave the non-issue related pure documentation ones in there? I normally wouldn't but given its complexity, debugging without the comments will be tough for me let along someone else? 

On the issue comments, don't think there is anything serious, just need to tidy up (and works as is) but will have tidied up on next commit.

> - Keep basic routes in original namespace (up for debate if you can convince me otherwise :D)

I'll amend to overload as suggested and test the override has no bugs (I had split them as subRoute is little different across two ... but that actually shouldn't make any difference really).

> - Add to the README a bit of a description what the TokenRouter is (as in how it works under the covers) and  when a user should opt in for it instead of the normal routes (basically make it more clear for less experienced users which to pick).

Will do, won't get too technical but explain basic radix node tree traversal & hand rolled parsers
> - Perhaps make it mandatory/optional for a user to configure a 404 handler for the TokenRouter, as I can imagine that many people will want to be able to customise the 404 response with performance routing as well

OK, I can add an initial first argument to `router` before route list that is the "`not found`" handler to run if not matched in the tree.
> - Make helper functions such as rtrn, between, etc. private or perhaps move into a sub module of TokenRouter so that these functions do not accidentally override the same named functions by a user who uses TokenRouter.

Oversight on my part, yes will re-review and ensure all internal helpers set to `private` as should be, and most are already
> - Restore all the metadata of the Giraffe.fsproj which has been removed in this PR

Oops, missed that, not sure how that happened, yeah may re-checkout master and then and re-create/copy TokenRouter & Tests file to make sure properly re-aligned 
> - Fix the last broken unit test :)

I don't know how this is broken though!? I would expect the result of the test to be "blah blah" not "blah%20blah" as we have used NSubstitute and not passed it through any middleware that would have re-encoded the space to a `%20` ... I mean I can Hack and pass through a url encode function but that's just as contrived as changing the initial expected result of "blah%20blah". How is the `%20` getting in there in the original tests without server on UNIT tests?

Thanks for the preliminary review and good spot on metadata, will have updates committed in next few days ... Just guidance on url encoding test is all I need final clarification on.

dustinmoris:
Comments in the code are fine and I totally agree, when the code is complex
then comments are always desired.

Re unit test I think something has changed in the unit testing framework
and I had to fix the same test on develop as well, but don't remember it
from the top of my head now. I think it has nothing to do with your
implementation so don't worry about it for now and I'll try to look into it
again to figure out what it was.

I'll be back checking on issues by end of this week again, hope you've got
everything you need until then!

Many thanks!

PR #106, opened 2017-10-09
Only get sub path if it is in the path

JonCanning:
When an exception is thrown in a subroute, and this middleware is used:

[ExceptionHandlerMiddleware](https://github.com/aspnet/Diagnostics/blob/rel/2.0.1/src/Microsoft.AspNetCore.Diagnostics/ExceptionHandler/ExceptionHandlerMiddleware.cs#L59)

`context.Request.Path` is set to `_options.ExceptionHandlingPath`, so `System.IndexOutOfRangeException: Index was outside the bounds of the array` was thrown on the subsequent call to `getPath`

### DIFF
--- a/src/Giraffe/HttpHandlers.fs
+++ b/src/Giraffe/HttpHandlers.fs
@@ -46,9 +46,9 @@ let private getSavedSubPath (ctx : HttpContext) =
 
 let private getPath (ctx : HttpContext) =
     match getSavedSubPath ctx with
-    | Some p -> ctx.Request.Path.ToString().[p.Length..]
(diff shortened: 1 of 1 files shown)


gerardtoconnor:
good catch, looks good to me :+1:

dustinmoris:
Hi, thanks for the fix! I pulled your changes locally to fix the broken unit test in AppVeyor, but now that the develop branch is up to date with your fix I can't merge your PR anymore, because it has already been merged to develop. GitHub doesn't pick it up this time, because this PR was against master. I will close it but as you can see it's been applied to develop.

PR #110, opened 2017-10-12
Discarded trailing slashes in the path when binding to a route.

sgryt:
For paths that end with a slash, the current implementation interprets that slash as being part of the value of the the last segment in the path. This is not what one would expect, given that slashes are path segment delimiters.

Incidentally, If the binding of the last member is done to a non-string property on a type, the binding throws an exception.

This PR (attempts to, at least) change that, so the binding becomes robust with respect to trailing slashes.

I hope you find this useful - otherwise, please let me know :)

### DIFF
--- a/src/Giraffe/HttpHandlers.fs
+++ b/src/Giraffe/HttpHandlers.fs
@@ -213,13 +213,14 @@ let routeCif (path : StringFormat<_, 'T>) (routeHandler : 'T -> HttpHandler) : H
             | Some args -> routeHandler args next ctx
 
 /// Filters an incoming HTTP request based on the request path (case insensitive).
+/// The filter ignores any trailing slashes on the path.
 /// The parameters from the string will be used to create an instance of 'T
 /// and subsequently passed into the supplied routeHandler.
 let routeBind<'T> (route: string) (routeHandler : 'T -> HttpHandler) : HttpHandler =
     fun (next : HttpFunc) (ctx : HttpContext) ->
         let pattern = route.Replace("{", "(?<").Replace("}", ">.+)") |> sprintf "^%s$"
         let regex = Regex(pattern, RegexOptions.IgnoreCase)
-        let mtch = regex.Match ctx.Request.Path.Value
+        let mtch = regex.Match (ctx.Request.Path.Value.TrimEnd('/'))
         match mtch.Success with
         | true ->
             let groups = mtch.Groups
--- a/tests/Giraffe.Tests/HttpHandlerTests.fs
+++ b/tests/Giraffe.Tests/HttpHandlerTests.fs
@@ -1352,3 +1352,37 @@ let ``GET "/{foo}/{bar}" returns Hello World``() =
             let body = getBody ctx
             Assert.Equal("Hello 1 f40580b1-d55b-4fe2-b6fb-ca4f90749a9d", body)
     }
+
+[<Fact>]
+let ``GET "/{foo}/{bar}/" ignores a trailing slash``() =
+    let ctx = Substitute.For<HttpContext>()
+    let app = GET >=> routeBind<RouteBind> "/{foo}/{bar}" (fun m -> sprintf "%s %i" m.Foo m.Bar |> text)
+    ctx.Request.Method.ReturnsForAnyArgs "GET" |> ignore
+    ctx.Request.Path.ReturnsForAnyArgs (PathString("/Hello/1/")) |> ignore
+    ctx.Response.Body <- new MemoryStream()
+    task {
+        let! result = app next ctx
+
+        match result with
+        | None     -> assertFail "It was expected that the result would be Hello 1"
+        | Some ctx ->
+            let body = getBody ctx
+            Assert.Equal("Hello 1", body)
+    }
+
+[<Fact>]
+let ``GET "/{foo}/{bar}////" ignores trailing slashes``() =
+    let ctx = Substitute.For<HttpContext>()
+    let app = GET >=> routeBind<RouteBind> "/{foo}/{bar}" (fun m -> sprintf "%s %i" m.Foo m.Bar |> text)
+    ctx.Request.Method.ReturnsForAnyArgs "GET" |> ignore
+    ctx.Request.Path.ReturnsForAnyArgs (PathString("/Hello/1////")) |> ignore
+    ctx.Response.Body <- new MemoryStream()
+    task {
+        let! result = app next ctx
+
+        match result with
+        | None     -> assertFail "It was expected that the result would be Hello 1"
+        | Some ctx ->
+            let body = getBody ctx
+            Assert.Equal("Hello 1", body)
+    }


dustinmoris:
Hi, thanks for looking into this and providing this PR.

I was actually looking into this at the time when I wrote that handler and my research has made me believe that in theory a URL which ends with a trailing slash is different than a URL without and both URLs could (according to the HTTP spec) refer to a different resource and therefore I feel that Giraffe should treat it this way too.

Have a look at some of the following discussions:
- https://webmasters.stackexchange.com/questions/2498/does-it-make-a-difference-if-your-url-path-ends-in-a-trailing-slash-or-not
- https://webmasters.googleblog.com/2010/04/to-slash-or-not-to-slash.html

However, I agree that most people would want their web server to not care and perhaps a better solution would be to add an option to the route handlers or add additional handlers where a trailing slash will be ignored. What do you think?

sgryt:
Hi Dustin,

Thanks for the swift response! 

I completely agree with your interpretation about the potential difference that a terminating slash may make to a web server, when determining which resource is being requested. But that is a slightly different concern than extracting path segment values, which is what I attempt to address with this PR.

In the context of 'dissecting' a path, slashes must always be considered delimiters, and never a part of the segment values (see https://tools.ietf.org/html/rfc3986#section-3.3). So I believe that in the context of binding route values, one must ignore terminating slashes, in order for the parser to not extract invalid values on the last path segment value.

I hope this clarifies my reasoning behind this PR?

gerardtoconnor:
From my perspective, it's on both parties to follow the standard so the server/dev should not have to hack the logic to catch exceptions. if you think the trailing slash should not be included, then leave it off the pattern match, if client is sending in slashes, fix the client... there is nothing I hate more then APIs that "fix" what I'm trying to do such that I'm pulling my hair out looking through source code trying to figure the bug... because that's what it is when it is following the requested logic operations  the user has requested. 

Also saying '/' are always arg delimiter is not correct in the real world regardless of standards (that most dont follow), query strings use ?,=,& and that's parsing args, a distinction without a difference.

I appreciate some my be having difficulty binding due to this issue but it's easily fixed on client/server side without hacking a correct implementation. This speciality can be invoked locally in app if needed.

sgryt:
Hi Gerard,

I am slightly confused by the combination of these parts of your comment: ’then leave it off the pattern match’ and ’without hacking a correct implementation’. How would I, as a user of Giraffe, change the pattern match without changing the implementation? 

About different delimiters: For the context of this function (routeBind), which only deals with the path segments (and not the query string or the fragment), there is only one delimiter: The slash. 
The current implementation does not treat these consistently, as only a terminating slash will be extracted as part of the segment value. All other slashes are not included in the extracted values. 

I don’t know of other server-side libraries/frameworks that treat terminating slashes the way that Giraffe currently does (again, in the context of parsing values out of path segments), but is sounds like you have come across such ones – could you provide a few pointers/links?

Best,
Mikkel

dustinmoris:
Hey guys,

thanks for your patience. I was thinking about what you both said and I can totally see how both sides make sense. I was playing around with it myself and came up with the following thinking/suggestion:

Overall I think the web server should not remove trailing slashes, because I think sooner or later someone might hit another edge case where a user had set up multiple routes which could be very similar, e.g.:

```fsharp
routeBind<RouteBind> "/{foo}/{bar}"
routeBind<RouteBind> "/{foo}/{bar}/{blah}"
routeBind<RouteBind> "/{foo}/{bar}/{blah}/{sth}"
```

and then I think it wouldn't be obvious what should happen in a case of `http://server.com/foo/bar///` anymore.

Some users might expect to map to the first route and others probably to map to one of the other two, therefore I think this should be configured by a user itself.

The route parameter is nothing more than a regex at the end of the day and therefore a user could set up a route like this:

```fsharp
routeBind<RouteBind> "/{foo}/{bar}/{blah}(/*)" // allow 0 or more trailing slashes
routeBind<RouteBind> "/{foo}/{bar}/{blah}(/?)" // allow 0 or 1 trailing slash
etc.
```

This would allow every application to decide for itself how to handle trailing slashes.

Only problem with this is that the current implementation of `routeBind` doesn't allow this regex to work, because the `.` which is used is too greedy and swallows the forward slashes as well.

A simple fix would be to change line 220 in `HttpHandlers.fs` to:

```fsharp
let pattern = route.Replace("{", "(?<").Replace("}", ">[^/\n]+)") |> sprintf "^%s$"
```

The change basically swaps the `.` with `[^/\n]` which means "every character except \n and a forward slash".

What do you both think of this solution?

gerardtoconnor:
@dustinmoris I don't mind really, whatever you prefer and maintains consistency while giving users the flexibility they need as I appreciate @sgryt should be able to impliment matching however he likes for his application's needs.

Just to point out on consistency:

- Trimming slashes would be different behaviour to how `routef` binds and damage consistency
- routeBind only binds strings so on the original point stating *"if the binding of the last member is done to a non-string property on a type, the binding throws an exception"*, seems impossible in current string-only implementation.
- string is a hybrid immutable ref type so doing this trim is allocating and copying new string (char []) everytime which on a high throughput server is just waste when could just not read/parse a trailing char instead. (not a concern for most I admit)

I don't use `routeBind` myself but hope to implement an upgraded version down the line once generative TP are solid on core such that one would not need to define a binding type (TP generates from string) and can bind all parseable types eg (`route<"/{foo:s}/{bar:i}/{fun}">.Bind(fun v -> printfn "%s : %i : %s" v.foo v.bar v.fun ...)`), not just string, also would not use JObject or Regex as these are slow, same reason erasing TP is not an option for parsing. 

`routeBind` in its current form is a nice convenience implementation and best we can do on .net core for now but would be nice to use TP in future if we can.

sgryt:
@dustinmoris Sounds fine to me 👍  I'm not a regex expert, but if you are certain that it never backtracks, I can't see why it would not be a stable solution.

I had not considered a scenario with more than one terminating slash (empty path segments). Probably because I have never come across a use case for such a path, but I certainly can't say that I have seen it all :)

@gerardtoconnor Below is a case that makes the current `routeBind` implemetation throw an exception during object instantiation. This is what lead me to look into the implementation in the first place. And given that I can't control if a client adds a terminating slash, and that it would be great not to have to register all routes twice (with and without a terminating slash, and in the correct sequence), I thought it would be useful to make the implementation a bit easier to work with:

```
[<CLIMutable>]
type Foo = { Name : string; Value : int }

HTTP GET "/fooName/1/"

 Giraffe.HttpHandlers.routeBind<Foo> "/{name}/{value}"
```

Exception: 
```
Newtonsoft.Json.JsonReaderException: Could not convert string to integer: 1/. Path 'value'.
   at Newtonsoft.Json.JsonReader.ReadInt32String(String s)
   at Newtonsoft.Json.JsonReader.ReadAsInt32()
   at Newtonsoft.Json.JsonReader.ReadForType(JsonContract contract, Boolean hasConverter)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.PopulateObject(Object newObject, JsonReader reader, JsonObjectContract contract, JsonProperty member, String id)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateObject(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerMember, Object existingValue)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.Deserialize(JsonReader reader, Type objectType, Boolean checkAdditionalContent)
   at Newtonsoft.Json.JsonSerializer.DeserializeInternal(JsonReader reader, Type objectType)
   at Newtonsoft.Json.Linq.JToken.ToObject(Type objectType, JsonSerializer jsonSerializer)
   at Newtonsoft.Json.Linq.JToken.ToObject(Type objectType)
   at Newtonsoft.Json.Linq.JToken.ToObject[T]()
   at Giraffe.HttpHandlers.routeBind[T](String route, FSharpFunc`2 routeHandler, FSharpFunc`2 next, HttpContext ctx)
   at UnitTests.Tests.routeTests@70.Invoke(Unit _arg1)
   at Expecto.Impl.execTestAsync@859-1.Invoke(Unit unitVar)
   at Microsoft.FSharp.Control.AsyncBuilderImpl.callA@839.Invoke(AsyncParams`1 args)
```

gerardtoconnor:
Cool I see, implicit conversion of sliced string into JSON.net is throwing conv exception.

I guess as you already know, your client caller is not following route format but I'm guessing you cannot control that if you need this work around. hopefully the regex extension will fix this for you.

dustinmoris:
@sgryt I've made the discussed changes and pushed to develop: https://github.com/dustinmoris/Giraffe/commit/885075f0540d6849475fa725cc2e7e1c44704556

Perhaps pull the latest dev branch and see if this fix fully satisfies your use case before I merge into master and create a new release.

sgryt:
@dustinmoris That seems to do the trick 👍 FWIW, my use case is described rather accurately by the following 2 unit tests, which both pass with this change:
```
[<Fact>]
let ``GET "/{foo}/{bar}/{id}/" returns Hello 3 f40580b1-d55b-4fe2-b6fb-ca4f90749a9d``() =
    let ctx = Substitute.For<HttpContext>()
    let app = GET >=> routeBind<RouteBind> "/{foo}/{bar}/{id}(/?)" (fun m -> sprintf "%s %i %O" m.Foo m.Bar m.Id |> text)
    ctx.Request.Method.ReturnsForAnyArgs "GET" |> ignore
    ctx.Request.Path.ReturnsForAnyArgs (PathString("/Hello/3/f40580b1-d55b-4fe2-b6fb-ca4f90749a9d/")) |> ignore
    ctx.Response.Body <- new MemoryStream()
    task {
        let! result = app next ctx

        match result with
        | None     -> assertFail "It was expected that the result would be Hello 3 1f40580b1-d55b-4fe2-b6fb-ca4f90749a9d"
        | Some ctx ->
            let body = getBody ctx
            Assert.Equal("Hello 3 f40580b1-d55b-4fe2-b6fb-ca4f90749a9d", body)
    }

[<Fact>]
let ``GET "/{foo}/{bar}/{id}" returns Hello 4 f40580b1-d55b-4fe2-b6fb-ca4f90749a9d``() =
    let ctx = Substitute.For<HttpContext>()
    let app = GET >=> routeBind<RouteBind> "/{foo}/{bar}/{id}(/?)" (fun m -> sprintf "%s %i %O" m.Foo m.Bar m.Id |> text)
    ctx.Request.Method.ReturnsForAnyArgs "GET" |> ignore
    ctx.Request.Path.ReturnsForAnyArgs (PathString("/Hello/4/f40580b1-d55b-4fe2-b6fb-ca4f90749a9d")) |> ignore
    ctx.Response.Body <- new MemoryStream()
    task {
        let! result = app next ctx

        match result with
        | None     -> assertFail "It was expected that the result would be Hello 4 1f40580b1-d55b-4fe2-b6fb-ca4f90749a9d"
        | Some ctx ->
            let body = getBody ctx
            Assert.Equal("Hello 4 f40580b1-d55b-4fe2-b6fb-ca4f90749a9d", body)
    }
```

sgryt:
@dustinmoris No worries :)

nojaf:
Are we missing some docs in the README for this?

dustinmoris:
@nojaf Yes, I'll add some!

PR #111, opened 2017-10-13
Fixed some typos in an example page

alinnert:
The CSS property `display` has no value `inline-box`. I think you meant `inline-block`.


dustinmoris:
Great catch! Thank you :)

PR #114, opened 2017-10-19
TokenRouter

gerardtoconnor:
Continuing from the previous PR I closed due to metadata being messed with, please see updates in line with updates requested, happy to go over any last bits. 

I separated the routers parsers out into a separate file to reduce clutter.

I amended `blah blah` test to what I expect and now passing but let me know if the url encoding inside unit testing is an issue.


dustinmoris:
Hey, thanks for the quick changes! I'll merge this PR for now as it looks good and as I said before the next time I'll get time to pick things up will be Sat/Sun this week. Gives you time to do any more tweaks and then I can create another release with the TokenRouter feature and even start thinking about a first RC release.

PR #118, opened 2017-10-27
UseGiraffeErrorHandler chaining support

draganjovanovic1:
UseGiraffeErrorHandler ApplicationBuilder extension modified (ignore removed) to support chaining since it should probably never be the last middleware in the pipeline.

```fsharp
app.UseCors(configureCors)
    .UseGiraffeErrorHandler(errorHandler)
    .UseAuthentication()
    .UseGiraffe webApp
```
vs
```fsharp
app.UseCors configureCors |> ignore
app.UseGiraffeErrorHandler errorHandler
app.UseAuthentication() |> ignore
app.UseGiraffe webApp
```


dustinmoris:
Nice, yes that makes sense. I wasn't thinking about chaining when I wrote the middleware, but I think your PR makes a lot more sense and is more incline with ASP.NET Core 👍

Issue #113, opened 2017-10-18
Format of writing responses (nested HttpHandlers)


gerardtoconnor:
I have noticed since the adoption of the new continuation format it has made the writing of some responses a little awkward looking and very evident in recent [MS blog on Giraffe](https://blogs.msdn.microsoft.com/dotnet/2017/09/26/build-a-web-service-with-f-and-net-core-2-0/)

Issue is around needing access to ctx and not being able to do ETA reduction of child arguments of internal HttpHandlers (nested vs pipelined), eg from blog:
```fsharp
let handleLunchFilter (next: HttpFunc) (ctx: HttpContext) =
    let filter = ctx.BindQueryString<LunchFilter>()
    let lunchSpots = LunchAccess.getLunches filter
    json lunchSpots next ctx                  // Inception style HttpHandler within a HttpHandler
```
Given the fact that writing JSON / XML / TEXT to the response closes out the pipeline in nearly all cases, we can introduce context extension methods to not only allow simpler (more intuative) writing of JSON etc but provide the short circuit (Task.FromResult Some ctx) such that handler could instead be written
```fsharp
let handleLunchFilter  _  ctx : HttpHandler =
    let filter = ctx.BindQueryString<LunchFilter>()
    let lunchSpots = LunchAccess.getLunches filter
    ctx.WriteJson(lunchSpots)                  // WriteJson : obj -> Task<HttpContext option>
```

It would not be too much less characters but 

- users get AutoComplete Intellesense for `ctx.WriteXXX()` 
- not having to deal with the less obvious appending of `next ctx` to apply to `HttpHandlers` 
- the user will get an error if they try to write after `.WriteJson` as it is already returning

The exiting HttpHandlers would still remain as they are useful for composed pipeline use vs nested use as displayed above.

I am happy to do PR for this provided everyone happy that it is not providing too much choice / polluting the HttpContext methods.

dustinmoris:
Great suggestion, a PR for this would be nice, thank you!

AlbertoDePena:
@gerardtoconnor 

Any update on the PR? =)

gerardtoconnor:
Apologies, manic week, will put PR in tomorrow

AlbertoDePena:
No worries, thanks =)

gerardtoconnor:
Merged in PR #119

PR #119, opened 2017-10-29
Added Nested Response Writing Methods

gerardtoconnor:
Added 3 common extension methods to context for writing json, xml & text so that users can do nested writes instead of using pipeline functions. #113

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@ The old NuGet package has been unlisted and will no longer receive any updates. 
     - [text](#text)
     - [json](#json)
     - [xml](#xml)
+    - [Nested Response Writing](#nested-response-writing)
     - [negotiate](#negotiate)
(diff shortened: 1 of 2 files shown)


gerardtoconnor:
seems to be an issue with travis, config error with powershell

> $ powershell ./build.ps1 -Release
powershell: command not found
The command "powershell ./build.ps1 -Release" exited with 127.

dustinmoris:
Thanks for this PR. I'll fix the Powershell issue now according to [this solution](https://github.com/PowerShell/PowerShell/issues/5225#issuecomment-339554138). One thing which I will probably rename is `WriteXML` to `WriteXml` as I've already started to write `Xml` with a lower m and l in other places and because I think this matches more to what MS has also done in the standard lib (e.g. `XmlWriter`), otherwise looks great!

PR #127, opened 2017-11-01
Fix template referencing wrong packages

vtquan:
Fix problem with the template using outdated Giraffe version 0.1.0-beta-110 and nonexistent Giraffe.Razor version 0.1.0-beta-200


dustinmoris:
Thank you, this was a stupid mistake of mine!

Issue #123, opened 2017-10-30
Lightweight policy based authorization


draganjovanovic1:
I was analyzing ASP.NET new approach with policy based authorization. It looks nice but is somehow too verbose. However, it gave me some nice idea. 

Example:
```fsharp
let mustBeJohn =
    requiresAuthPolicy (fun user -> user.HasClaim (ClaimTypes.Name, "John")) accessDenied

let app =
    choose [
        route "/ping" >=> text "pong"
        route "/john-only" >=> mustBeJohn >=> text "Hi John."
    ]
```

The requiresAuthPolicy function is higher order function which accepts ClaimsPrincipal -> bool policy and can be used as a base function for all the existing auth functions.

Example:
```fsharp
let requiresRole (role : string) (authFailedHandler : HttpHandler) : HttpHandler =
    requiresAuthPolicy
        (fun user -> user.IsInRole role)
        authFailedHandler
```

@dustinmoris, if you are interested I have PR with sample and the docs ready.

dustinmoris:
Hey, this sounds good. Feel free to send a PR for this :)

dustinmoris:
Thanks for the PR, has been merged and will be released with the next version!

PR #128, opened 2017-11-02
Lightweight policy based authorization

draganjovanovic1:
#123

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,7 @@ The old NuGet package has been unlisted and will no longer receive any updates. 
     - [mustAccept](#mustaccept)
     - [challenge](#challenge)
     - [signOff](#signoff)
+    - [requiresAuthPolicy](#requiresauthpolicy)
     - [requiresAuthentication](#requiresauthentication)
     - [requiresRole](#requiresrole)
     - [requiresRoleOf](#requiresroleof)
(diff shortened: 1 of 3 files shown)


nojaf:
@draganjovanovic1 @dustinmoris aren't we missing unit tests here?

dustinmoris:
Yeah I'll add it before releasing!

On 2 Nov 2017 09:21, "Florian Verdonck" <notifications@github.com> wrote:

@draganjovanovic1 <https://github.com/draganjovanovic1> @dustinmoris
<https://github.com/dustinmoris> aren't we missing unit tests here?

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/dustinmoris/Giraffe/pull/128#issuecomment-341404587>,
or mute the thread

draganjovanovic1:
@nojaf, yes you are right, also all the existing auth handler are missing tests. @dustinmoris, I can make pull request with unit tests. Let's agree not to do the same job :) 

BTW Have you thought about using FsCheck?

dustinmoris:
Hey, sure if you want to provide a PR that would be awesome. Regarding FsCheck how do you think it would add value to the current tests? Is it not more suitable for property based testing which is good for algorithms but maybe less for very specific authentication scenarios? I might be misunderstanding the library though so if you have more experience and could explain the benefit of it that it would be great.

forki:
Property based testing is introduced for testing of algorithms and
protocols. Because these are usually the hardest things to tests. But it
turns out they are valuable for most other things as well.

Am 06.11.2017 04:07 schrieb "Dustin Moris Gorski" <notifications@github.com
>:

Hey, sure if you want to provide a PR that would be awesome. Regarding
FsCheck how do you think it would add value to the current tests? Is it not
more suitable for property based testing which is good for algorithms but
maybe less for very specific authentication scenarios? I might be
misunderstanding the library though so if you have more experience and
could explain the benefit of it that it would be great.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/dustinmoris/Giraffe/pull/128#issuecomment-342035081>,
or mute the thread

PR #129, opened 2017-11-03
Modified regex generation so that tokens are identified at the end of the request path.

KerryRJ:
Regex generation only worked for paths that started and ended with tokens (/{foo}/{bar}/{id} and not for paths /api/{foo}/{bar}/{id}.


dustinmoris:
Great, thank you for the fix!

PR #131, opened 2017-11-05
Use camel-case for property values returned as JSON

dsincl12:
Returned JSON should be camel-cased as this is the standard format and what is expected.

### DIFF
--- a/src/Giraffe/Common.fs
+++ b/src/Giraffe/Common.fs
@@ -6,6 +6,7 @@ open System.Text
 open System.Xml
 open System.Xml.Serialization
 open Newtonsoft.Json
+open Newtonsoft.Json.Serialization
 
 /// ---------------------------
 /// Helper functions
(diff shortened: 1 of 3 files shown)


dustinmoris:
Hi, thanks for this PR. Even though this is a breaking change I think it makes sense to go with the default JSON convention, which is also what (as you stated) [MVC has done in its latest version](https://github.com/aspnet/Announcements/issues/194).

I'll make sure to change the docs accordingly too and make a clear note of this change in the upcoming release notes. This is also slightly related to #97.

KerryRJ:
Will this not completely block the use of JsonConvert.DefaultSettings?

dustinmoris:
@KerryRJ Hmm yes it will in it's current form. The plan in #97 is to allow a user to pass in custom json settings if the default is not desired. I'll probably have to implement this before releasing this PR.

PR #134, opened 2017-11-06
Added %g for Guids as a format placeholder.

KerryRJ:


JonCanning:
Could the hyphens be optional?

JonCanning:
Sorry, I should have said Welcome first before blurting out my random thoughts

😟

nojaf:
I believe so, I was having the same thought.

```fsharp
> let myGuid = System.Guid.NewGuid().ToString("N");;
val myGuid : string = "c9f69caf68a349ed83bb7e9ddfd22c24"

> System.Guid.TryParse(myGuid);;
val it : bool * System.Guid = (true, c9f69caf-68a3-49ed-83bb-7e9ddfd22c24)
```

KerryRJ:
I don't know Regex enough to make all dashes all or none.

nojaf:
I was thinking `([0-9A-Fa-f]{8}\-[0-9A-Fa-f]{4}\-[0-9A-Fa-f]{4}\-[0-9A-Fa-f]{4}\-[0-9A-Fa-f]{12})|([0-9A-Fa-f]{8}[0-9A-Fa-f]{4}[0-9A-Fa-f]{4}[0-9A-Fa-f]{4}[0-9A-Fa-f]{12})` but no regex hero either.

KerryRJ:
Thank you for the regex.

JonCanning:
Could the second match be [0-9A-Fa-f]{32} ?

KerryRJ:
Trying to rebase and messing things up.

gerardtoconnor:
I have one small question with implementation in that `settings` is being added to the end of most of the curried functions with the value being applied first. Usually, the most frequently applied parameter is applied latest such that partial application can be used to create an instance of the partial/shortened function so it once again has the original signature but we only create one base closure that requires the value to be applied on.

Example in short:
Would have expected this
``` 
let inline serializeJson x (settings: JsonSerializerSettings) = JsonConvert.SerializeObject(x, settings)
```
To look more like:
```
let inline serializeJson (settings: JsonSerializerSettings) x = JsonConvert.SerializeObject(x, settings)
```
does it have to be in the order you have done or is it possible to swap argument curry order?

KerryRJ:
I did originally use curry order but couldn’t get piping to work in one of the existing functions/methods, probably due to my inexperience.
Let me take another crack at it and swap them around.

Kind Regards

Kerry Jones    Owning Member of Info-Overdrive cc
m. +27(0)612403717 | o. +27(0)333902367
e. kerry@iodrive.co.za<mailto:kerry@iodrive.co.za>
s. Muskham Farm F1, Chief Mhlabunzima Rd., Pietermaritzburg, 3201

Issue #136, opened 2017-11-07
BindJson crashes Kestrel on Azure


dsincl12:
Took me hours to figure this one out but I finally found the error: "Concurrent reads or writes not allowed." That's what Kestrel crashes with on Azure when BindJson is called more than once. I had severe issues with my "real" application and in the end just copy and pasted submitCar from the BindJson example and got the same error.

The first call to an endpoint with BindJson always goes through but subsequent calls returns "502 - Web server received an invalid response while acting as a gateway or proxy server." or "HTTP Error 502.3 - Bad Gateway The specified CGI application encountered an error and the server terminated the process." This is just Kestrel crashing with the error mentioned above.

Locally everything works just fine and I have no issues at all.

dawedawe:
I can reproduce this. For me, this also only happens on Azure.
Here's a stack trace.

2017-11-08 06:01:44.423 +00:00 [Error] Giraffe.Middleware.GiraffeErrorHandlerMiddleware: An unhandled exception has occurred while executing the request.

System.NullReferenceException: Object reference not set to an instance of an object.

   at System.IO.StreamReader.IsPreamble()

   at System.IO.StreamReader.<ReadBufferAsync>d__65.MoveNext()

--- End of stack trace from previous location where exception was thrown ---

   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()

   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)

   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)

   at System.Runtime.CompilerServices.ConfiguredTaskAwaitable`1.ConfiguredTaskAwaiter.GetResult()

   at System.IO.StreamReader.<ReadToEndAsyncInternal>d__61.MoveNext()

--- End of stack trace from previous location where exception was thrown ---

   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()

   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)

   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)

   at Giraffe.HttpContextExtensions.HttpContext-BindJson@62-1.Invoke(Unit unitVar0)

   at Giraffe.TaskBuilder.StepStateMachine`1.nextAwaitable()

[...]

PR #137, opened 2017-11-08
Use synchronous .ReadToEnd() in .ReadBodyFromRequest() fixes issue #136

dsincl12:
May be a quick fix but for now, but it works.


dustinmoris:
Hey, thanks for this fix and reporting the issue. Seems like that @gerardtoconnor fix by deserialising from the stream directly has fixed this so I'm closing this PR without merge.

PR #139, opened 2017-11-15
Use Path.Combine in htmlFile to specify the path

forki:


forki:
Also I have a question about this.
env.ContentRootPath where does this come from? I was setting the web root to another directory but contentrootpath still points to current directory

gerardtoconnor:
Hi,

This can be set/changed in the WebHostBuilder config:

example `main` 
```fsharp
let main argv =                                
    WebHostBuilder()
        .UseKestrel()
        .UseContentRoot(Directory.GetCurrentDirectory())  // << ContentRoot set here to Currrent Directory
        .ConfigureServices(Action<IServiceCollection> configureServices)
        .ConfigureLogging(Action<ILoggerFactory> configureLogging)
        .Build()
        .Run()
    0
```

forki:
what's the difference to usewebroot?

gerardtoconnor:
content root is generally your root folder with code etc, webroot usually your "wwwroot" folder that is publically accessible and contains html, css etc, the folder `on the web` vs the `server content`

This shorty piece helps
[getting-the-web-root-path-and-the-content-root-path-in-asp-net-core](https://blog.mariusschulz.com/2016/05/22/getting-the-web-root-path-and-the-content-root-path-in-asp-net-core)

forki:
ok so this should be relative to wwwroot, right?

gerardtoconnor:
I think both default to project root `.\` for `ContentRootPath` and root\wwwroot `.\wwwroot` and if you want to override change either, you need to use absolute paths for both such that.

There may be a better way of getting current relative root but all examples I have seen just use `Directory.GetCurrentDirectory()` and I have not tested path strings like `.\` etc. what you set ContentRoot to will also be where complied bin folder goes, just to bear in mind,

```fsharp
let main argv =                                
    WebHostBuilder()
        .UseKestrel()
        .UseContentRoot(Directory.GetCurrentDirectory())
        .UseWebRoot(Directory.GetCurrentDirectory() + @"\wwwroot\")
        .ConfigureServices(Action<IServiceCollection> configureServices)
        //.ConfigureLogging(Action<ILoggerFactory> configureLogging)
        .Build()
        .Run()
    0
```

forki:
anyway I changed it to allow full path. that should make it work for more scenarios

gerardtoconnor:
Thanks for the PR, good to get fresh set of eyes on the design.

dustinmoris:
Great change, thank you!

PR #141, opened 2017-11-15
fix StringFormat to PrintfFormat as bug in type inference

gerardtoconnor:
This is urgent fix needed as users reporting routef handler not inferring types correctly and defaulting to obj when uncertain. there is a bug in StringFormat<_,'T> but changing to PrintfFormat<_,_,_,_,'T> gets rid of this inference bug

### DIFF
--- a/src/Giraffe/FormatExpressions.fs
+++ b/src/Giraffe/FormatExpressions.fs
@@ -38,7 +38,7 @@ let convertToRegexPatternAndFormatChars (formatString : string) =
     |> convert
     |> (fun (pattern, formatChars) -> sprintf "^%s$" pattern, formatChars)
 
-let tryMatchInput (format : StringFormat<_, 'T>) (input : string) (ignoreCase : bool) =
+let tryMatchInput (format : PrintfFormat<_,_,_,_, 'T>) (input : string) (ignoreCase : bool) =
     try
         let pattern, formatChars = 
             format.Value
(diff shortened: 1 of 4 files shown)


gerardtoconnor:
I've also added the update on ctx.BindJson<'T>() to use json stream reader and avoid creating unneeded intermediate string (parses json stream directly to container)

gerardtoconnor:
Also added URL decoding to parsed strings to ensure %tags converted to correct chars

dustinmoris:
Thanks, I'll try to release this and the latest changes shortly

dustinmoris:
@gerardtoconnor Quick question, what is the bug with StringFormat? Has it been logged somewhere in the relevant repo?

gerardtoconnor:
Hey, It's a bug in that the type info is lost on the closure abbreviation and the types between the StringFormat and the HttpHandler were not being enforced and free be anything (based on what it was inferring in hander). It's kind of a bug but as we were using StringFormat in an unexpected way, I'm not sure it will be fixed when `PrintfFormat` does work. It's located in FSharp.Core, Printf module. Do you feel the urge to use StringFormat again?

dustinmoris:
I was just working on an application where it would have been convenient to be `StringFormat` because I wanted to use the same format for a route and later in another function with `sprintf` and I couldn't pass the `PrintfFormat` into `sprintf` but it was not a big deal, was just making me think again why we had to change it and if we think this is a long term fix on our side or if it is temporary until it gets fixed in FSharp.Core.

Issue #140, opened 2017-11-15
Suave RequestErrors


forki:
Hey,

in my current project we were heavily using Suave:RequestErrors module. I can't find a replacement here so I started to port things. Are you interested in a PR whichg would add it as Giraffe functions?

gerardtoconnor:
This sounds nice, and I'd be all for it, saves user having to remember the error codes

forki:
see #142

dustinmoris:
Merged and will go out with the next release.

PR #142, opened 2017-11-15
Porting most HTTP Status helpers from suave

forki:
as discussed in #140

### DIFF
--- a/src/Giraffe/HttpStatus.fs
+++ b/src/Giraffe/HttpStatus.fs
@@ -1,0 +1,130 @@
+namespace Giraffe
+open HttpHandlers
+
+module Intermediate =
+    let CONTINUE next ctx  = (setStatusCode 100 >=> setBody [||]) next ctx
+    
+    let SWITCHING_PROTO next ctx = (setStatusCode 101 >=> setBody [||]) next ctx
+
+
+// 2xx
+module Successful =
+
+    let ok s = setStatusCode 200 >=> s
+
+    let OK s = ok (text s)
+
+    let created s = setStatusCode 201 >=> s
+
+    let CREATED s = created (text s)
+
+    let accepted s = setStatusCode 202 >=> s
+
+    let ACCEPTED s = accepted (text s)
+
+// 3xx
+module Redirection =
+
+
+    let moved_permanently location =
+        setStatusCode 301 
+        >=> setHttpHeader "Location" location
+        >=> setBody [||]
+    
+    let MOVED_PERMANENTLY location = moved_permanently location
+
+    let found location =
+        setStatusCode 302
+        >=> setHttpHeader "Location" location
+        >=> setBody [||]
+
+    let FOUND location = found location
+
+    let redirect url =
+        let str = sprintf """<html><body><a href=\"%s\">%s</a></body></html>""" url "Object moved temporarily -- see URI list"
+
+        setStatusCode 302
+        >=> setHttpHeader "Location" url
+        >=> setHttpHeader "Content-Type" "text/html; charset=utf-8"
+        >=> setBodyAsString str
+
+// 4xx
+module RequestErrors = 
+
+    let bad_request s = setStatusCode 400 >=> s
+
+    let BAD_REQUEST message = bad_request (text message)
+
+
+    /// 401: see http://stackoverflow.com/questions/3297048/403-forbidden-vs-401-unauthorized-http-responses/12675357
+    let unauthorized s =
+        setStatusCode 401 
+        >=> setHttpHeader "WWW-Authenticate" "Basic realm=\"protected\"" 
+        >=> s
+
+    let UNAUTHORIZED message = unauthorized (text message)
+ 
+    let forbidden s = setStatusCode 403 >=> s
+
+    let FORBIDDEN message = forbidden (text message)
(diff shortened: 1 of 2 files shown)


forki:
If people are interested then I will port over the docs. But not today

src/Giraffe/HttpStatus.fs
```diff
@@ -0,0 +1,130 @@
+namespace Giraffe
+open HttpHandlers
+
+module Intermediate =
+    let CONTINUE next ctx  = (setStatusCode 100 >=> setBody [||]) next ctx
```
gerardtoconnor: can remove `next ctx` on args and tail

src/Giraffe/HttpStatus.fs
```diff
@@ -0,0 +1,130 @@
...
+    
+    let SWITCHING_PROTO next ctx = (setStatusCode 101 >=> setBody [||]) next ctx
```
gerardtoconnor: can remove next ctx on args and tail

dustinmoris:
Hi, thanks for this PR - I think it's useful and will review and merge the PR after I had a good night sleep and release it with the next version - it's currently 4:44AM so apologies that I didn't include this one in tonight's 0.1.0-beta-300 release :)

tforkmann:
Hi,

this is a great pull request!

Would be nice to have that implemented!

AlbertoDePena:
Fantastic!

src/Giraffe/HttpStatus.fs
```diff
@@ -0,0 +1,130 @@
...
+    let BAD_REQUEST message = bad_request (text message)
+
+
+    /// 401: see http://stackoverflow.com/questions/3297048/403-forbidden-vs-401-unauthorized-http-responses/12675357
+    let unauthorized s =
+        setStatusCode 401 
+        >=> setHttpHeader "WWW-Authenticate" "Basic realm=\"protected\"" 
+        >=> s
+
+    let UNAUTHORIZED message = unauthorized (text message)
```
dustinmoris: Commenting on this line, but this applies to all other status codes as well:

One thing which I personally don't like is to prescribe a text message by default - this was one of the reasons why I rarely used the predefined status handlers in Suave for example, because if I had an API which mostly spoke JSON then I also wanted to return error messages and other things in JSON.

I think in Giraffe an alternative would be to perhaps use `negotiate` instead of `text` by default, so that the passed in parameter `s` could be of any type as well. What do you guys think?

forki: That's why we have the lower case versions.

dustinmoris: Yes I agree, and I thought additionally it would maybe make sense to change `text` to `negotiate` so there's these two options:
- lowercase version where a user can force a specific response
- uppercase version where the framework will reply in the client's preferred mime type by default

I think this would make the uppercase version overall more useful. I am happy with the rest of this PR and otherwise ready to merge, but would like to change this unless there's a good reason to keep as text - any thoughts?

forki: The reason is that we actually specify error messages here. And this keeps it compatible with suave so that porting apps is easy

dustinmoris: Ok, from my own experience I know that many clients have specific requirements though, they will send HTTP headers where they request a response in a set of understood mime types and when the error response comes back in plain text it's often no good. The fact that the message is a string is no big deal, because if a client sends `text/plain` then they will get the response this way, but if it requests `application/json` then the response will be at least formatted into a valid JSON like `"Some error message"` instead of just `Some error message`.

In terms of Suave compatibility I appreciate that it might be a bit easier for people to port, but I don't intend to lure Suave users away from Suave and therefore make a migration very easy. As we already established in another issue I think Suave has it's own user base and application for very good reasons and Giraffe for other reasons and I don't want to follow every model from Suave only because things have been implemented it in a certain way there. I am happy for Giraffe to do things differently where I see advantages like in this one. In all honestly Suave is also further evolving and perhaps this is even a change which they might like to adopt, so nothing is set in stone and things can change or be different if you ask me.

forki: I guess then you need to change it. /shrug

forki: I really think you are making a mistake, but if you don't want to change it you I can probably change it sometime next week or so.

dustinmoris: What is the mistake though? I'd be happy to discuss it objectively to see the pros of text vs negotiate.

forki: The mistake is to use own design for established things. And negotiate is perfectly fine to use with lower case versions. That's why they are there.

Anyway there are more important things, and it's your project.

dustinmoris: Ok, thank you for the PR anyway, I'll merge it and then change it to 'negotiate' probably because at the moment I feel this is the right way for Giraffe, but I appreciate your thoughts on this.

I think I might see it slightly differently to you as I don't consider the way Suave has done it as the "established" way, because it's literally only one project which has done it in a way rather than the way - but I'm happy to be shown more frameworks which follow the same principle.

Thanks for the PR 👍

forki:
What is the status here? Is it released. Would love to get rid of my own copy of this from my code base

PR #145, opened 2017-11-17
Added guid placeholder to route paths

KerryRJ:
Message up the previous pull request due to inexperience.
Recovered and am initiating a new pull request based on #134.


dustinmoris:
Thanks for this PR, I'll check it out tomorrow and hope to merge it then!

src/Giraffe/TokenParsers.fs
```diff
@@ -92,6 +92,11 @@ let private floatParse (path:string) ipos fpos =
     | '+' -> go (ipos + 1)
     | _ -> go (ipos)
 
+let private guidParse (path:string) ipos fpos =
```
gerardtoconnor: The performance parser is not supposed to allocate (substring), working solution but I have already built a rolling parser 2x-3x faster then below so we can use that if you don't mind?

gerardtoconnor:
@KerryRJ  how are you getting on with this, do you need any assistance ?

gerardtoconnor:
if you want to swap my guid parser in before merge will save me having to do separate commit after, parser code is just:
```
let private guidMap = [|3;2;1;0;5;4;7;6;8;9;10;11;12;13;14;15|]
let private guidParse (path:string) ipos fpos =
    let byteAry = Array.zeroCreate<byte>(16)
    let mutable bytePos = 0
    let mutable byteCur = 0uy
    let mutable atHead = true
    let rec go pos =
        if path.[pos] = '-' then // skip over '-' chars
            if pos < fpos then go (pos + 1) else failure
        else 
            let cv =  byte path.[pos]
            
            let value =
                if  cv >= byte '0' then
                    if cv <= byte '9' then cv - byte '0'
                    elif cv >= byte 'A' then
                        if cv <= byte 'F' then cv - byte 'A' + 10uy
                        elif cv >= byte 'a' then
                            if cv <= byte 'f' then cv - byte 'a' + 10uy
                            else 255uy
                        else 255uy
                    else 255uy
                else 255uy                                        

            if value = 255uy then
                failure
            else
                if atHead then
                    byteCur <- value <<< 4
                    atHead <- false
                    go (pos + 1)   // continue iter
                else
                    byteAry.[guidMap.[bytePos]] <- byteCur ||| value
                    if bytePos = 15 then
                        Guid(byteAry) |> box |> rtrn
                    else
                        byteCur <- 0uy
                        atHead <- true
                        bytePos <- bytePos + 1
                        go (pos + 1)   // continue iter
    //Start Parse
    go (ipos)
```
Its a little verbose but intentionally so as it is 2x-3x faster then substring/Guid.TryParse.
Only allocation we do is the internal constructor array for guid that needs to be allocated anyway so this is about as lightweight as we'll get it.

KerryRJ:
I'll add your parser in tomorrow @gerardtoconnor.
Thank you.

JonCanning:
That guid parser has blown my mind

gerardtoconnor:
Thanks @JonCanning :blush: as you may have guessed, I am a little obsessed on performance & throughput but shows f# can be a lot faster then native .net implementations if compromise a bit on FP

PR #146, opened 2017-11-17
Use custom JsonSerializerSettings and not JsonConvert.DefaultSettings for serializing and deserializing JSON.

KerryRJ:


dustinmoris:
Thanks for that!

PR #147, opened 2017-11-18
Changed base tag to voidTag

nojaf:
According to mdn,the `base` tag is an empty one.
See https://developer.mozilla.org/nl/docs/Web/HTML/Element/base


dustinmoris:
Great spot, thanks as always!

Issue #178, opened 2017-12-16
Make JSON serializer configurable


alfonsogarciacaro:
Reference https://github.com/SAFE-Stack/SAFE-BookStore/pull/255#issuecomment-352194907

Some users may have particular needs in JSON serialization. For example, when interacting with Fable apps you need to [pass the Fable.JsonConverter](http://fable.io/docs/interacting.html#json-serialization). This means you cannot use the default http handlers in Giraffe and they need to be customized. Two alternatives have been proposed to fix but they both have inconveniences:

- Write a custom `negotiate` function, but users need to remember to use it every time (like `Successful.ok (myCustomNegotiate promos) next ctx`) and the compiler won't help them when they don't.
- Create a Giraffe.Fable package, but it looks too big an effort to create a full package to override **all** Giraffe handlers only in order to pass a custom converter to JSON serialization settings.

Instead of those :point_up: would you be OK with making the [defaultJsonSerializerSettings](https://github.com/giraffe-fsharp/Giraffe/blob/60c29e2e8d21886b853066fef2492b8450f12a87/src/Giraffe/Common.fs#L40) configurable? And if so, what would be best way to do it (besides making that field mutable)?

dustinmoris:
Hmm... I agree that it is not ideal at the moment and the more often I read your issue and think about other similar issues in the past I agree that we could have done much better with the current implementation.

The first thought which comes to my mind would be to use ASP.NET Core's DI model to solve this issue in a more elegant way.

For example, we could change the current `json` http handler to the following:

```fsharp
let json (dataObj : obj) : HttpHandler =
    fun (next : HttpFunc) (ctx : HttpContext) ->
        let settings = 
            match ctx.GetService<JsonSerializerSettings>() with
            | null -> defaultJsonSerializerSettings
            | x    -> x
        (setHttpHeader "Content-Type" "application/json"
        >=> setBodyAsString (serializeJson settings dataObj)) next ctx
```

Then in your ASP.NET Core startup you can optionally decide to configure custom `JsoSerializerSettins`:

```fsharp
let configureServices (services : IServiceCollection) =
    services.AddSingleton<JsonSerializerSettings>(
        JsonSerializerSettings(ContractResolver = CamelCasePropertyNamesContractResolver())) |> ignore
```

With this we could probably also get rid of `customJson` and all overloads in the `HttpContext` extension methods.

[...]

PR #179, opened 2017-12-18
Make NegotiationRules and UnacceptableHandler configurable

forki:


forki:
this is another way to solve https://github.com/giraffe-fsharp/Giraffe/issues/178

src/Giraffe/HttpHandlers.fs
```diff
@@ -377,6 +377,35 @@ let negotiateWith (negotiationRules    : IDictionary<string, obj -> HttpHandler>
                         |> fun mimeType -> negotiationRules.[mimeType.MediaType.Value]
                         |> fun handler  -> handler responseObj next ctx
 
+
+
+let DefaultNegotiationRules =
```
gerardtoconnor: why are we creating a list, than adding to dictionary? would the prior syntax of `dict [ ... ]` not be better as module values (this list) will be persisted when not needed after dictionary construction? not too big a deal and might be to allow further additions  (as dict is read-only)? in this case we should probably define `DefaultNegotiationRules` as scoped variable in `NegotiationRules` and additionally, we should pass the count to the Dictionary constructor to avoid re-sizing of known dataset?

forki:
changed it to dict

forki:
and changed it back to System.Collections.Generic.Dictionary because dict makes it readonly

src/Giraffe/HttpHandlers.fs
```diff
@@ -387,23 +417,7 @@ let negotiateWith (negotiationRules    : IDictionary<string, obj -> HttpHandler>
...
-        ])
-        // Default unacceptable HttpHandler
-        (fun (next : HttpFunc) (ctx : HttpContext) ->
-            (setStatusCode 406
-            >=> ((ctx.Request.Headers.["Accept"]).ToString()
-                |> sprintf "%s is unacceptable by the server."
-                |> text)) next ctx)
-        // response object
-        responseObj
+    negotiateWith NegotiationRules UnacceptableHandler responseObj
```
gerardtoconnor: Are we able to use ETA reduction here such that 
`let negotiate = negotiateWith NegotiationRules UnacceptableHandler`
so the function is pre-applied and doesn't need to wrap closure each time at runtime?

Sorry very minor tweak but we try to pre-build as much as possible to leave minimal to runtime.

forki: that is compiler decision if you ask me.

dustinmoris:
@forki & @gerardtoconnor: Do you both think that using a globally mutable variable for settings such as this and JSON and potentially a lot more stuff in the future is the best/preferred approach in F# and ASP.NET Core? (genuine question)

Here are my thoughts: I cannot think of why this should be a problem and I think this approach is sort of in direct competition of the DI approach which was discussed in #178 and I would just want to think about all implications before setting on one or the other. If you think this is the best way of doing it then I am happy to go with this approach, but then I think we should keep it consistent. I don't want to end up in a scenario where parts of the library get configured via services and the other part via global variables.

Also if we go with global variables, then I think it would be nice to create a single config type, e.g. `Giraffe.Settings` or `Giraffe.Config`, etc. where a user can set things like `Giraffe.Config.Negotiation.Rules` and `Giraffe.Config.Negotiation.UnacceptableHandler`, `Giraffe.Config.Serialization.Json.Settings`, etc.

This would make it a bit friendlier as in a user wouldn't have to search for all possible mutable settings variables. They could just use Intellisense and browse the `Giraffe.Config` object.

What do you think? Great work by the way... every time I feel like Giraffe is getting more and more stable someone comes with another great suggestion how to make it better. Unfortunately still many breaking changes but definitely worth it!

forki:
I'm not saying it#s the best solution ever. But yes I think it's better than DI

gerardtoconnor:
I think DI should be there for compatibility with asp.net core but we can do better for all Giraffe specific functionality. looking up keys, and casting obj on every other call seems bizarre to me when we know at compile time where everything fits into place.

The only issue with having a mutable state obj, is that we need to ensure that some `checking` function, that runs near end, is checking for values and injecting default where empty unless we create obj with defaults first and then have them overwritten when desired. Usually, I would hate an approach of creating & allocating an obj only to have replaced, but in this case, its only a cost at startup so probably worth it.

dustinmoris:
If we go with mutable then I would want to set defaults and protect them behind functions, so that it is physically impossible to set null and all code can always rely on using them without extra null checks.

When I think about global variables then I don't like that testing a Giraffe application will become a lot harder. If I'd want to change the negotiation rules between two tests how would I do it with globals?

forki:
it's in no way harder than DI. It's just a different setup step.

gerardtoconnor:
Yeah, I agree with @forki , DI is doing the same thing, but with more abstraction, less control etc. 

We could create config type with properties for each config with null checked setters. A bit of boiler plate but it's a clean performant solution with null checks only one on a setting change (typically at startup).

dustinmoris:
How though?

gerardtoconnor:
@dustinmoris possible example would be
```f#
module Giraffe =
    let mutable private jsonConverter = DefaultSettings()
    let mutable private anotherConfig = DefaultOtherConfig()

    type Config =
        static member JsonConverter with get () = jsonConverter and set v = if not(isNull v) then jsonConverter <- v
        static member AnotherConfig with get () = anotherConfig and set v = if not(isNull v) then anotherConfig <- v
```
usage:
```f#
Giraffe.Config.JsonConverter <- CustomFableConveter
```
Tests would then use same access pattern.

gerardtoconnor:
On the above we have complete type safety without DI dict/cast as well as nice intelisense on the static config type so users can explore available overridable config settings ... DI is a bit of a black hole in that regard

dustinmoris:
@gerardtoconnor but then ALL tests in that assembly would be forced to use `CustomFableConveter` right? What if I want to have one test using `CustomFableConveter` and another test using `CustomOtherConveter`?

forki:
Just put config on top of the test or test list
. Like you would do with injection. There is no difference

Am 18.12.2017 18:55 schrieb "Dustin Moris Gorski" <notifications@github.com

gerardtoconnor:
As @forki said on each test, at beginning put:
```f#
Giraffe.Config.JsonConverter <- ConveterForTest
```
But in real life, all tests in a group will be using a specific converter, not on a test by test basis, for fable tests, would be in a fable test file where we can set the converter at the start of the file. If you were relying on copping and changing a converter through DI, would have the same problem.

forki:
@gerardtoconnor I think we would probably register a INegotiationConfig. Then it's just one lookup per request. Plus one lookup in the config itself

forki:
Ok this is not in the spirit of the giraffe framework. Let's inject a json converter instead.

PR #180, opened 2017-12-18
Set encoding for ReadBodyFromRequestAsync

forki:


forki:
Can you please release this one as a patch. Currently it seems to create null reference exceptions. I have a repro to check if things are fixed.

dustinmoris:
Sure will do!

PR #181, opened 2017-12-18
Small cleanup in negotiateWith

forki:


src/Giraffe/HttpHandlers.fs
```diff
@@ -351,29 +351,27 @@ let renderHtml (document : XmlNode) : HttpHandler =
...
                   : HttpHandler =
     fun (next : HttpFunc) (ctx : HttpContext) ->
         (ctx.Request.GetTypedHeaders()).Accept
         |> fun acceptedMimeTypes ->
             match isNull acceptedMimeTypes || acceptedMimeTypes.Count = 0 with
-            | true  ->
-                negotiationRules.Keys
+            | true ->
+                negotiationRules
                 |> Seq.head
```
gerardtoconnor: Not a reflection on your tidy but of original code, I think this is a bug waiting to happen as IDictionary is an interface on any Map/Dictionary and worked to date given we are using Generic Dictionary that happens to enumerate in order added, if a Map was passed, the first value would be different. I think it makes more sense to have default handler as an argument to function rather then a magic entry in a dictionary that feels quite hacked and wont work properly in many cases

forki: you can't pass a map.

gerardtoconnor:
Original code had a lot of temporary structures, further to my suggestion of explicit default handler outside the dictionary, I have done an imperative version that allocates a lot less.

```f#
let negotiateWith (negotiationRules    : IDictionary<string, obj -> HttpHandler>)
                  (unacceptableHandler : HttpHandler)
                  (defaultHandler : obj -> HttpHandler)
                  (responseObj         : obj)
                  : HttpHandler =
        (ctx.Request.GetTypedHeaders()).Accept
        |> fun acceptedMimeTypes ->
            match isNull acceptedMimeTypes || acceptedMimeTypes.Count = 0 with
            | true  -> defaultHandler responseObj next ctx
            | false ->
                let mutable mimeType = Unchecked.defaultof<_>
                let mutable curQuality = 0.
                for x in acceptedMimeTypes do
                    if negotiationRules.ContainsKey x.MediaType.Value then
                        if x.Quality.HasValue then 
                            if x.Quality.Value > curQuality then
                                curQuality <- x.Quality.Value
                                mimeType <- x
                if Object.ReferenceEquals(null,mimeType) then
                    unacceptableHandler next ctx 
                else
                    negotiationRules.[mimeType.MediaType.Value] responseObj next ctx```

File: src/Giraffe/HttpHandlers.fs (hunk 1 above)
Comment (gerardtoconnor) [reply src/Giraffe/HttpHandlers.fs]: But Map has an IDictionary Interface implemented, what stops function accepting it?

Comment (forki) [reply src/Giraffe/HttpHandlers.fs]: in any case the fallback needs to be made explicit

**Comment by forki** [comment]
there must be a better way than this ;-) maybe it allocates fewer lists, but it will be hard to maintain

**Comment by gerardtoconnor** [comment]
😄 Not for me, less elegant but yes, for many might be harder to maintain ... I think it is sometimes ok to go more verbose in platform/library code where performance is a feature, but imperative perf code aside, I do think pulling the default handler out of the dictionary is better and a hack in its current form.

**Comment by forki** [comment]
it's not about imperative, but things Object.ReferenceEquals(null,mimeType) are just weird. did you check with `isNull`?

**Comment by forki** [comment]
also it contains a bug. if no negotiationRules has a quality then you return unacceptableHandler

**Comment by gerardtoconnor** [comment]
Fair enough, was just a quick bash out of code apologies. I used Object.ReferenceEquals being pedantic as `isNull` is not inline and causes extra call, on the negotiationRules, good spot on quality issue, can be fixed easily and I'm sure there are one or two other things I overlooked, was just offering baseline for less allocaty process.

**Comment by forki** [comment]
isNull will be compiler optimized. It's fine to use. My point was more: it's harder to maintain for everyone. Even original author ;-)

**Comment by gerardtoconnor** [comment]
hacky fix but let's forget about it as difficult to maintain. Would like to know @dustinmoris thoughts on `defaultHandler` instead of special case dictionary entry.
```f#
let negotiateWith (negotiationRules    : IDictionary<string, obj -> HttpHandler>)
                  (unacceptableHandler : HttpHandler)
                  (defaultHandler : obj -> HttpHandler)
                  (responseObj         : obj)
                  : HttpHandler =
        (ctx.Request.GetTypedHeaders()).Accept
        |> fun acceptedMimeTypes ->
            match isNull acceptedMimeTypes || acceptedMimeTypes.Count = 0 with
            | true  -> defaultHandler responseObj next ctx
            | false ->
                let mutable mimeType = Unchecked.defaultof<_>
                let mutable curQuality = -1.
                for x in acceptedMimeTypes do
                    if negotiationRules.ContainsKey x.MediaType.Value then
                        if x.Quality.HasValue then 
                            if x.Quality.Value > curQuality then
                                curQuality <- x.Quality.Value
                                mimeType <- x
                        else
                            if curQuality < 0. then
                                curQuality <- 0.
                                mimeType <- x
                if isNull mimeType then
                    unacceptableHandler next ctx 
                else
                    negotiationRules.[mimeType.MediaType.Value] responseObj next ctx
```

forki:
@gerardtoconnor what about de71571?

gerardtoconnor:
Better as now the order of IDictionary cannot mess up expected default ... but if we're passing parameter anyway ... why not save the lookup and call the function directly!? See what @dustinmoris thinks and happy to go with whatever you guys think best as I'm being overly performance focused.

forki:
I want to pass the key because I want to override what "*/*" means in that other PR

forki:
ok applied the loop version. can you please review

src/Giraffe/HttpHandlers.fs
```diff
@@ -351,31 +351,34 @@ let renderHtml (document : XmlNode) : HttpHandler =
 /// Example:
 /// let negotiationRules = dict [ "application/json", json; "application/xml" , xml ]
 /// `json` and `xml` are both the respective default HttpHandler functions in this example.
-let negotiateWith (negotiationRules    : IDictionary<string, obj -> HttpHandler>)
+let negotiateWith (negotiationRules    : #IDictionary<string, obj -> HttpHandler>)
```
dustinmoris: What does the `#` do here? (noob question)

gerardtoconnor: it allows implicit acceptance of interface without need to cast the type to the interface (basically)

forki: it means "flexible types" and makes F# accept "more" dictionaries than before. I generally see it as good habit, while it probably does not change anything here. Type is already fixed to string * obj -> HTTPHandler. I revert that change

dustinmoris: Thanks for the explanation!

src/Giraffe/HttpHandlers.fs
```diff
@@ -351,31 +351,34 @@ let renderHtml (document : XmlNode) : HttpHandler =
...
                   (unacceptableHandler : HttpHandler)
+                  (defaultMimeType     : string)
```
dustinmoris: Personally I think this is redundant/duplicate information. The default mime type is only relevant when a client has no preference whatsoever, in that case the server already has a dict of supported mime types and a user has to pass it into `negotiateWith`, so they might as well set the dict in preferred order (even though it doesn't matter to the client).

dustinmoris: As in what I am trying to say is that I think that a default mime type as a concept doesn't make sense, because if a client doesn't care then they don't care and so why should the server care which one to serve. The response is only for the client and if the client doesn't care, then the server might as well just pick a random one and instead of random one I think taking the first from the list has at least an expected behaviour.

forki: yes the problem is: there is no first. If you don't care and random is what you want then we can revert that

dustinmoris: You mean a dict has no concept of first? Yes, I'd be happy to revert and pick a random, I'd also be happy to use a collection type which has a first/last. I think the only things I'd want is that if I make the same request from a client which has no Accept header that the server will at least always return a response in the same mimeType.

forki: done

forki: yes it's still predictable. In the dictionary case it will select first in list. In map case it will select the one with smallest key (alphabetical order)

dustinmoris:
Hey, so first I agree that my original code wasn't the best probably and I liked what @forki had written before https://github.com/giraffe-fsharp/Giraffe/pull/181/commits/5893d4d3b7be584d660d374f61b59633949beefe because the code got a lot more readable. Anyone who would want to know what `negotiateWith` or `negotiate` for that matter does would be very easily to find out by looking at the source code. The use of mutable variables feels to me less functional and idiomatic and apart from making it harder to read I don't know what other implications it might have. I feel like we are trying to micro optimise the code by avoiding to write idiomatic F# which could have benefited from current or future compiler optimisation by the F# SDK. If you guys feel strongly about it then I can cave it, but maybe add two lines of comments above which explain to new people who look at the code what's going on.

src/Giraffe/HttpHandlers.fs
gerardtoconnor: yeah that's what I was trying to get at earlier, it will be consistent but different depending on dict/map type

forki:
I think @gerardtoconnor has a point here. That function is called A LOT and we should not create too many intermediate collections. I think the current version is a good compromise and still okish. I added two line comment

dustinmoris:
Ok, thanks for the improvements, looks good to me.

gerardtoconnor:
@dustinmoris no I fully get where you're coming from and I would prefer if we could write functionally and have same performance but unfortunately, in the prior version we were creating multiple temporary data structures per call, and the function called a lot so it can make sense in these situations to sacrifice some elegance for perf ... I have had a lot of performance bottlenecks in the past due to too may temp data structures, it's why Java Streams and Nessos Streams are so much faster, they combine all the operations of a pipeline into single iteration per value, no temp structures.

PR #182, opened 2017-12-18
WIP: Add WebSockets lib

forki:
Not sure where it belongs.

- [ ] - discuss if it should go in separate package
- [x] - XML document the functions
- [ ] - document the sockets lib in readme
- [x] - add tests
- [ ] - add samples


TheAngryByrd:
Might we worth adding some tests too using `TestServer` like

```Fsharp
use server = new TestServer(WebHostBuilder().Configure(fun app -> configure app))
let wsClient = server.CreateWebSocketClient()
```

TheAngryByrd:
There also might be some value comparing and contrasting this implementation with https://gist.github.com/TheAngryByrd/22b5df5a308f9f96f959980d0a42f7cd

For instance we probably should rely on [WebSocketReceiveResult.EndOfMessage](https://msdn.microsoft.com/en-us/library/system.net.websockets.websocketreceiveresult.endofmessage(v=vs.110).aspx) 
 in the receive part and buffer into a stream rather than hoping the message size is big enough to handle the incoming text data.

forki:
@TheAngryByrd please send PRs to my fork

forki:
@@TheAngryByrd  I made receive into a loop. I would appreciate some tests

forki:
@theangrybyrd thanks for working with me together on this. I'm sure we can create something cool out of it.

tests/Giraffe.Tests/Giraffe.Tests.fsproj
```diff
@@ -12,12 +12,17 @@
     <PackageReference Include="Microsoft.AspNetCore.Hosting" Version="2.0.*" />
     <PackageReference Include="Microsoft.AspNetCore.TestHost" Version="2.0.*" />
     <PackageReference Include="Microsoft.AspNetCore.StaticFiles" Version="2.0.*" />
+
+    <PackageReference Include="Microsoft.AspNetCore.WebSockets" Version="2.0.*" />
```
forki: This is so dangerous. It should be pinned or converted to paket. Floating versions are such an anti pattern. Especially together with transitive deps.

forki:
@TheAngryByrd I'm now using TokenRouter to configure the server. And we have a broadcast test

enricosada:
@forki check https://github.com/fsharp/FsAutoComplete/blob/c5cde65a47879f92bf262bb1ea414df5edd809d4/test/FsAutoComplete.IntegrationTests/TestHelpers.fsx#L194 you need to test also EndOfMessage property

enricosada:
Another good idea ihmo (found checking ws with suave in FSAC) will be implement the autobahn test suite to test the server http://websockets.github.io/ws/autobahn/servers/ to test the implementation ( link is the node ws test report ).

forki:
@TheAngryByrd @gerardtoconnor for some reason the tests with n=100 / 1000 are going into timeout. Any ideas why it is soooo slow?

forki:
ok I found the issue

forki:
@theangrybyrd could you please review this and comment on the current version?

src/Giraffe/WebSockets.fs
```diff
@@ -0,0 +1,182 @@
...
+                        if not cancellationToken.IsCancellationRequested then
+                            let webSocket = kv.Value.WebSocket
+                            if webSocket.State = WebSocketState.Open then
+                                do! webSocket.SendAsync(segment, WebSocketMessageType.Text, true, cancellationToken)
+                            else
+                                if webSocket.State = WebSocketState.Closed then
+                                    connections.TryRemove kv.Key |> ignore
+                    with
+                    | _ -> () })
+                |> Task.WhenAll
```
vasily-kirichenko: This does not limit parallelism. Is it OK?

src/Giraffe/WebSockets.fs
```diff
@@ -0,0 +1,182 @@
...
+                connections
+                |> Seq.map (fun kv -> task {
+                    try
+                        if not cancellationToken.IsCancellationRequested then
+                            let webSocket = kv.Value.WebSocket
+                            if webSocket.State = WebSocketState.Open then
+                                do! webSocket.SendAsync(segment, WebSocketMessageType.Text, true, cancellationToken)
+                            else
+                                if webSocket.State = WebSocketState.Closed then
+                                    connections.TryRemove kv.Key |> ignore
```
vasily-kirichenko: Removing element from a collection that's being enumerated... I don't know if it's OK for ConcurrentDictionary, but it's a strong code smell.

src/Giraffe/WebSockets.fs
```diff
@@ -0,0 +1,182 @@
...
+                            keepRunning <- true
+                            endOfMessage <- true
+                        | _ ->
+                            raise (NotImplementedException())
+
+            return keepRunning
+        }
+
+        member private this.RegisterClient<'Msg>(reference:WebSocketReference,connectedF: WebSocketReference -> Task<unit>,messageF,cancellationToken:CancellationToken) = task {
+            connections.AddOrUpdate(reference.ID, reference, Func<_,_,_>(fun _ _ -> reference)) |> ignore
```
vasily-kirichenko:
```fsharp
connections.AddOrUpdate(reference.ID, reference, fun _ _ -> reference) |> ignore
```

src/Giraffe/WebSockets.fs
```diff
@@ -0,0 +1,182 @@
...
+            do! connectedF reference
+            let mutable running = true
+            while running && not cancellationToken.IsCancellationRequested do
+                let! msg = this.Receive<'Msg>(reference,messageF,cancellationToken)
+                running <- msg
```
vasily-kirichenko: There are a lot of imperative code like this in this PR. Is it for performance reasons?

src/Giraffe/WebSockets.fs
```diff
@@ -0,0 +1,182 @@
...
+            | true, reference -> do! reference.CloseAsync()
+            | _ -> ()
+        }
+
+        /// Creates a new WebSocket connection and negotiates the subprotocol.
+        member this.CreateSocket(onConnected,onMessage,?webSocketID,?supportedProtocols:seq<WebSocketSubprotocol>,?cancellationToken) =
+            let negotiateSubProtocol(requestedSubProtocols,supportedProtocols:seq<WebSocketSubprotocol>) =
+                supportedProtocols
+                |> Seq.tryFind (fun (supported:WebSocketSubprotocol) ->
+                    requestedSubProtocols |> Seq.contains supported.Name)
```
vasily-kirichenko: O(N^2). Is it OK here?

src/Giraffe/WebSockets.fs
TheAngryByrd: I though task's can be tail call optimized https://github.com/giraffe-fsharp/Giraffe.Tasks/blob/master/src/Giraffe.Tasks/Tasks.fs#L229

src/Giraffe/WebSockets.fs
gerardtoconnor: The Sequence is generating tasks from connections enumeration to allow next socket to be called before the last task finished, as would be the case with for loop in task CE, Task.WhenAll ensures returned method task marks completed when all sockets have been completed (rather than fire&forget), where do you see the parallelism bottleneck?

src/Giraffe/WebSockets.fs
gerardtoconnor: I agree, this would definitely fail on the normal dictionary but due to locks in Concurrent version, might be preventing usual error. Might be cleaner/safer to add a temp "IdsToRemove" list that is then enumerated after the initial run to remove.

src/Giraffe/WebSockets.fs
gerardtoconnor: This is an interesting one. Although it looks imperative, due to CE nested expressions it will end up not looking imperative at all in IL but importantly to do iterative work in Task CE, you need to use imperative loops of while/for as rec call would be creating and destroying the Async state machine etc every call and be a lot less efficient.

On tasks not being tail-call optimised, that is not entirely true, there is an initial unwrap needed as you've pointed but the rest of the chain is fine and can be TCO from a stack frame perspective, between functions of same Task type. The reason Task is not great for tail call recursion (which uses TCO) is that it leaves the whole linked chain of tasks on the heap and GC cannot collect so you get an OOM exception, but not a stack overflow exception (as would be the case if not able to TCO). 

I think the way Robert implemented state machine, he may have avoided this chaining issue (hence initial task wrap on mutating task state) but this will need testing as I think there is a good possibility we may get an OOM exception on this. @forki have you memory profiled at all?

src/Giraffe/WebSockets.fs
gerardtoconnor: is "negotiateSubProtocol" even used, I cannot see where fun it is used in the method scope? I think may have been a helper function but below you manually implement same code L171.

Given this is run on many requests and slow may need to think of a better structure, on 1000s of sockets would hurt perf. This would be a good candidate for creating a scoped Set value from seq (before fun returned so captured in closure)

No big deal but you can probably move "run" L159 out of HttpHandler, above into constructor and pass the 'next' 'ctx' args, so we are not creating that fun every time for no good reason?

src/Giraffe/WebSockets.fs
```diff
@@ -0,0 +1,182 @@
...
+    webSocketOptions
+
+/// Internal reference to a WebSocket. Includes WebSocketID and Subprotocol.
+type WebSocketReference = {
+    /// A reference to the WebSocket.
+    WebSocket : WebSocket
+    /// The selected subprotocol.
+    Subprotocol : WebSocketSubprotocol option
+    /// The internal ID of the WebSocket.
+    ID : string
```
gerardtoconnor: What is the benefit of using string when we are using strings of Guids, I imagine it is to allow easier transport of ID across wire?

src/Giraffe/WebSockets.fs
TheAngryByrd: Since these should be I/O tasks I don't think there will be a problem.

src/Giraffe/WebSockets.fs
TheAngryByrd: I guess it depends on if we want to let people choose their ids.  I could see people using a username from a jwt.

src/Giraffe/WebSockets.fs
forki: done

src/Giraffe/WebSockets.fs
forki: done

src/Giraffe/WebSockets.fs
forki: No I did not. So please advise how to rewrite

src/Giraffe/WebSockets.fs
forki: removed the duplication

src/Giraffe/WebSockets.fs
forki: yes we need to let people choose IDs

forki:
Incorporated all feedback

dustinmoris:
Hi, I don't like letting people wait, so I've [written an update on the Swagger PR](https://github.com/giraffe-fsharp/Giraffe/pull/218#issuecomment-365240525) and I was thinking of proposing the same here if that sounds good to you @forki ?

forki:
I'm fine with waiting for now.

dustinmoris:
Ok cool, just wanted to let everyone know that things are progressing every day and I'll eventually get to the bigger PRs as well. If something is urgent then there is that option of an early Alpha release.

cotyar:
Guys, 
For the ones who don't want to wait until everything is done properly I just created a branch merging both PRs together (just as a temporal solution)
[https://github.com/cotyar/Giraffe/tree/SwaggerAndWebsockets](https://github.com/cotyar/Giraffe/tree/SwaggerAndWebsockets)
Cheers

Banashek:
I had created a similar implementation based on [the asp docs](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/websockets) and [this blog post](https://radu-matei.com/blog/aspnet-core-websockets-middleware/). I like the PR here a bit more.

One thing I noticed in the current PR state is that there is no opportunity to add behavior when a socket is closed. This is useful for updating UI when a user has left the web app. If there was perhaps an overload of `onMessage` where `messageF` was of type `WebSocketReference -> string -> WebSocketMessageType -> Task<unit>` it could allow patching in behavior based on the message type by the user.

forki:
@banashek can you please send a pull request to my fork with your proposal? Let's work together on this to make it great

Nhowka:
I plan to create a version of [Elmish.Remoting](https://github.com/Nhowka/Elmish.Remoting) for Giraffe. Should I wait a little more for this PR to get merged?

cotyar:
No updates from Dustin for quite a while unfortunately. 
And the latest version of ws breaks the previous code contract unfortunately and isn't a dropin replacement :-(

I would love to see the remoting working over WebSockets for Giraffe!

forki:
This is still WIP since I'm still trying to figure out an idiomatic way

Nhowka:
Based on what you did until now, do you think it would be too much effort to replicate what Suave did?

cotyar:
I personally use 3 months old version (without latest commits) and it does the job pretty well for me, but I'm missing onClose event. The latest one has onClose but doesn't pass the ref into onOpen (as far as I remember)

cotyar:
Thank you @forki , much appreciated! 

Hope @dustinmoris will kindly merge it into the master and release soon together with the Swagger PR.

dustinmoris:
I wasn't actively working on this one myself, but I'm happy to put this
into a Giraffe.Sockets package when this work is done? Last thing I
remember it was still WIP?

forki:
My current WIP version is working fine in production for quite a while now.
Maybe we should just release it to a new package and see what users feel
about it

Dustin Moris Gorski <notifications@github.com> schrieb am Sa., 29. Sep.
2018, 13:30:

dustinmoris:
Sounds good to me! Do you think you can maybe also provide a bit of
documentation around it? Something that can go into the README of the
package repo? It doesn't have to be big, maybe look at TokenRouter as an
example or the Razor view engine package.

forki:
Sorry this completely wrong approach. I will come up with a new one

Issue #184, opened 2017-12-19
routef with %d doesn't work?


isaacabraham:
I have a `routef` route that looks like `%s/%s/%d`. It never finds a matching route at runtime but compiles. If I replace the route with `%s/%s/%i`, it works fine.

dustinmoris:
`%i` stands for integer and `%d` for `int64`, which is probably what is causing the issue. This is non standard, but because there is currently no format parameter for int64s and two format parameters for int32 (why?) I decided to use one for int64s, because I thought that this might be desired for some applications.

gerardtoconnor:
@dustinmoris not sure what we can do but most users seems to use '%d' by default for int and as a result get these issues continually raised. 

@isaacabraham do you normally always use `%d` over '%i' for int pattern?

dustinmoris:
Yes.. it was a 50/50 chance after all. Ideally I would want to get a format parameter for `int64`, something like `%l` (for long) and then remove either `%i` or `%d` because I don't even understand why there is two format strings for a single data type. I always thought that `%d` is confusing as it can be easily misinterpreted by someone less familiar with F# to stand for `DateTime` or `Decimal` and this was probably why I randomly used `%i` for `int` and `%d` for `int64`. :/

isaacabraham:
@gerardtoconnor yeah, I personally always use `%d`. Not sure if that's a good indicator of what's representative of the community though :-)

You could look to ameliorate this issue by making the routing engine go pop if it ever sees a `%d` route with guidance to move over to `%i`, and / or put some `///` comment that says the same thing?

gerardtoconnor:
yeah it's a tough one to handle. I think we need to build in a type checker at startup (once-off check) to ensure the types match up so users can get immediate feedback. I'll have a look at this.

vtquan:
When I first got started with Giraffe, I tried to use ``%d`` because I just come from the Suave Music Store tutorial which used ``%d`` and expecting it to carry over since the rest of the syntax was so familiar. I am sure that this is the case for a lot of people. Beside a type checker, I would like to see documentation on how to match on an int64 because without it, even though I know that ``%d`` is for int64, I just assume that they are treated the same but with ``http://localhost:5000/4294967297`` matching on ``%d`` which it doesn't

gerardtoconnor:
About to do PR now, error received at startup looks like:
```
[...]

PR #185, opened 2017-12-19
added parseValidate

gerardtoconnor:
To address an issue a number of users are having eg #184 , this validation will run at startup and warn the user of any invalid parse matching rather than it just failing to match at runtime for no apparent reason.

Test also added to cover common case of '%d' being expected to be a `int` rather than `int64`

@forki @isaacabraham

### DIFF
--- a/src/Giraffe/FormatExpressions.fs
+++ b/src/Giraffe/FormatExpressions.fs
@@ -82,16 +82,6 @@ let tryMatchInput (format : PrintfFormat<_,_,_,_, 'T>) (input : string) (ignoreC
     with
     | _ -> None
 
-let private matchDict = 
-    dict [
-        'b', typeof<bool>           // bool
-        'c', typeof<char>           // char
-        's', typeof<string>         // string
-        'i', typeof<int32>          // int
-        'd', typeof<int64>          // int64
-        'f', typeof<float>          // float
-        'O', typeof<System.Guid>    // guid
-]
 let parseValidate (format : PrintfFormat<_,_,_,_, 'T>) =
 
     let t = typeof<'T>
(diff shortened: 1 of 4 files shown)


src/Giraffe/FormatExpressions.fs
```diff
@@ -80,4 +80,57 @@ let tryMatchInput (format : PrintfFormat<_,_,_,_, 'T>) (input : string) (ignoreC
             :?> 'T
             |> Some
     with
-    | _ -> None
\ No newline at end of file
+    | _ -> None
+
+let matchDict =
```
dustinmoris: Can we make this guy private perhaps?

gerardtoconnor: I've made private, given these are only run at startup won't impact perf constructing on each call, I can even change to a match statement which is probably better as then literals.

gerardtoconnor:
removed the match dictionary and replaced with literal match

dustinmoris:
I think it would be nice to make the exception messages a bit more consistent so that they are easily recognisable once a user has encountered one of them.

Some repeating pattern like:

```
Format string error: Number of parameters (%i) does not match number of tuple variables (%i).

Format string error: Number of parameters (%i) does not match number of varaibles (%i).

Format string error: The parameter '%%%c' does not match expected type %s of tuple %s.

Format string error: Unkown validation error: %A [%i].
```

dustinmoris:
Sorry one last request, maybe we could name it `validateFormat`? I think the rest is good!

gerardtoconnor:
Out for dinner, will do later, feel free to do yourself if you can

isaacabraham:
Great to see this But I would make the error message more basic, something like: %d is not supported in routef. Use %i instead.

gerardtoconnor:
@isaacabraham that would work if `%d` was not supported, it is supported for int64 parsing so for that reason, along with '%O' being used for `Guid` we kind of need to flesh out the expected and received. Will try simplify though in line with @dustinmoris recommendation above.

@dustinmoris I have updated messages but on the main one:
```
Format string error: The parameter '%%%c' does not match expected type %s of tuple %s

Format string error: The parameter '%d' does not match expected type System.Int32 of tuple System.Int64
```
Did not read clear to me so tweaked with a printer to help on tuples

```
Format string error: Handler parse type ( _ , System.Int32 ) does not match expected type of parameter '%d' (System.Int64).
```
Printing out tuple structure is clearer for me but can tweak further if needed.

Issue #177, opened 2017-12-15
ReadBodyFromRequestAsync sometimes returns empty string even when body is not empty


cmeeren:
I'm experiencing a weird problem. When I use `ctx.ReadBodyFromRequestAsync()`, it will occasionally return empty string instead of the body I know for sure is there. I try the same request over and over again, and occasionally I get an empty string. See the images below - "ContentLength" indicates there is indeed content, but I get an empty string.

![contentlength](https://user-images.githubusercontent.com/7766733/34039724-f93cabea-e190-11e7-8405-a8faec4233fd.PNG)

![body](https://user-images.githubusercontent.com/7766733/34039726-f95a8cf0-e190-11e7-9aad-714395500ee7.PNG)

This seems to me a rather critical error.

gerardtoconnor:
Hi @cmeeren old friend 😄 
We have encountered fringe cases like this before where stream appears to be disposed of aggressively by asp.net core before properly read (or after an initial read).

In order to further look into this, can you provide any prior handlers in the pipeline before you try to read the body in the handler above?

It may be that something is triggering a read of body stream before this and disposing of the stream hence why header still showing content but body empty when you go to read.

cmeeren:
```f#
choose [
  ...
  route "/v4" >=> trackApiVersion "v4" >=> POST >=> handleV4
```

where `trackApiVersion` is defined like

```f#
let trackApiVersion (v:string) next ctx =
  trackEvent v
  next ctx
```

and `handleV4` calls `ReadBodyFromRequestAsync` on its first line.

cmeeren:
And as I said, this happens randomly, maybe 1 out of 10 times

cmeeren:
I also occasionally get a `NullReferenceException` on `ctx.ReadBodyFromRequestAsync()` instead of an empty body.

cmeeren:
Also note that I cannot reproduce the problem with `BindJsonAsync`. The reason I'm using `ReadBodyFromRequestAsync` and deserializing using `JsonConvert.DeserializeObject` is, as mentioned in #175, that my app and API communicates by sending shared, serialized objects (in an assembly referenced by both) back and forth, and I need to be sure that serialization and deserialization are compatible.

dustinmoris:
[...]

PR #186, opened 2017-12-19
Create StreamReader inside task CE

hvester:
I believe that this should fix issue [177](https://github.com/giraffe-fsharp/Giraffe/issues/177). If the StreamReader is created without using task CE then it is disposed when the method returns the task that
might not be completed.


dustinmoris:
Are you sure this is required? [It didn't have the task {} before](https://github.com/hvester/Giraffe/commit/ea395099c3d8bd88ab501ce224ed7705c23a9fac#diff-242b195e122369f466cfeb3a9810e4f3), but also I assigned the request body to a temporary variable, which might have made the difference.

Can you please try out [the NuGet package from the successful build](https://ci.appveyor.com/project/dustinmoris/giraffe/build/artifacts) before we merge and release another hotfix?

forki:
The problem is that the streamreader is disposed at the end of the method.
And we read later. The fix is good. And mine was useless. -500 and -510 are
both broken

Am 19.12.2017 20:03 schrieb "Dustin Moris Gorski" <notifications@github.com

forki:
But the good thing is we have a reproducible which repros 100 percent for
500 and 510. Do we can apply the fix and see if the hypothesis holds. And
we can't break something that is already broken ;-)

Am 19.12.2017 20:07 schrieb "Steffen Forkmann" <sforkmann@gmail.com>:

dustinmoris:
Ok, as soon as the checks finished I'll do another release! Thanks for the super quick fix!

gerardtoconnor:
Yeah surprised we didn't spot this sooner, the IDisposable interface is called as soon as initial reading Task is returned and the `use` variable looses scope, probably need to scan the codebase for any other cases of `use` outside task CE

dustinmoris:
merged into master, new release coming any minute.

@gerardtoconnor Good idea, I did a quick search and the only other usage of `use` was with non async methods (like in tests), so this was the only occurrence.

gerardtoconnor:
Can you review merge the parserValidator too while your at it so we can close related issues ?

dustinmoris:
Sure, will do!

forki:
Good news. It actually fixed the issue

Issue #138, opened 2017-11-11
Html module improvement


nojaf:
I already have the code for a PR but I'd like to have a discussion first whether people are on board with this.
Here goes:

The `Giraffe.XmlViewEngine` module contains helper functions for constructing html. It does really feel right to have this as part of the xml engine. That is why I propose we moved the html specific bits to a separate module `Giraffe.Html`.

So I moved all the stuff and added helper functions for html attributes as well.
See https://github.com/nojaf/Giraffe/blob/develop/src/Giraffe/Html.fs

I also re-exposed some stuff from the XmlViewEngine in this Html module so that you basically only need to add `open Giraffe.Html` to get everything.

```fsharp
type HtmlNode = XmlNode
let str = encodedText
let raw = rawText
let emptyText = XmlViewEngine.emptyText
let comment = XmlViewEngine.comment
```

The change from `encodedText` to `str` is to have more parity with [Fable.React](https://github.com/fable-compiler/fable-react/blob/master/src/Fable.React/Fable.Helpers.React.fs#L754). The [same reason](https://github.com/fable-compiler/fable-react/blob/master/src/Fable.React/Fable.Helpers.React.fs#L7) that `Attributes` are in a separate module.

Some elements and attributes have the same name. I listed the workaround [here](https://github.com/nojaf/Giraffe#renderhtml).

Check out the [Identity sample](https://github.com/nojaf/Giraffe/commit/7724d7b85b722492961aba5af3660432b11255da) to see before and after.

Please let me know what you think, I'm looking forward to send the PR 😉

dustinmoris:
Hi Florian,

thanks for the suggestion, I can see the reason for moving the HTML stuff into a separate module, but with the current suggestion I see the following problems:

- Name inconsistency: It would be `Giraffe.XmlViewEngine` vs `Giraffe.Html`. Maybe `Giraffe.Xml` and `Giraffe.Html` would be more consistent or `Giraffe.HtmlViewEngine`. `Giraffe.Xml` could be easily confused with generic XML serialization functions though, so probably wouldn't be a great name (and why I named it `...ViewEngine`)
- Code duplication: I would like to avoid having to declare certain things twice only because of a name change, especially when the underlying code is pretty much the same. I think it makes maintenance a lot easier and also less likely for a contributor to forget a change or feature which would apply to both - XML and HTML.
[...]

PR #187, opened 2017-12-21
Added attributes module

nojaf:
As originally discussed in https://github.com/giraffe-fsharp/Giraffe/issues/138 I'd like to add the html attributes. I've put them in a separate module but did add a prefix to clashing names.
I think it will be more practical to work with than module alias tricks.


dustinmoris:
Hey, thanks for this PR.

Do you think we could maybe slightly change the names of the attributs? What I don't like about the current names is that there is three different variations how one attribute might have been defined, but only one will work and a user won't know unless they have looked it up or tried it out:

- `class`
- ```class```
- `class_attr`

Ideally I would like to have one consistent way of declaring all attributes. Maybe we could prepend an underscore to all of them?

Example:

`_class`, `_data`, `_acceptCharset`, etc. 

Attributes like `accept-charset` I would therefore also prefer to write in camel case so that we don't have to use ```` part of the name. What do you think, would an additional underscore everywhere feel weird?

nojaf:
Good point, changed the names.

dustinmoris:
Nice thank you, I think this looks really good now! One last question, what do you think of adding the `[<AutoOpen>]` attribute above the `Attributes` module? This would have the effect that if someone opens `open Giraffe.GiraffeViewEngine` that they will also have the `Giraffe.GiraffeViewEngine.Attributes` open without having to write the second open statement.

Just to be clear, I am not suggesting that this is what we should do, just putting it out there for consideration and wanting to know if you think this would be useful? I've done this with other modules in a recent release. For example most of the default functionality can be opened with a single `open Giraffe` now, so maybe a single `open Giraffe.GiraffeViewEngine` would be enough here as well?

dustinmoris:
Also restarting the one failed Travis build.. looks like it was a hiccup with Travis which happens occasionally...

nojaf:
Hmm, I guess it does seem handy to have the `[<AutoOpen>]`. I'll send a PR.

PR #194, opened 2017-12-25
Add unsafe version of routef to Toke Router

Krzysztof-Cieslak:
Adds to Token Router version of `routef` that doesn't perform checking of format on the runtime.

### DIFF
--- a/src/Giraffe/TokenRouter.fs
+++ b/src/Giraffe/TokenRouter.fs
@@ -270,7 +270,7 @@ let route (path:string) (fn:HttpHandler) (root:Node) =
     Node.ExtendPath root path (fn |> HandlerMap |> End)
 
 ///**Description**
-/// Matches and parses url values from a route using `printf` string formatting, results are passes to function as tuple.
+/// Matches and parses url values from a route using `printf` string formatting, results are passes to function as tuple. Doesn't validate format on the runtime.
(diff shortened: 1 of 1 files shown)


dustinmoris:
@Krzysztof-Cieslak Thanks for the PR. What is the thinking behind having `routefUnsafe`? The way @gerardtoconnor added the validation it gets validated only once during startup and never again afterwards, so every time a http request comes in it will not do the validation again. If we go with this change I'd suggest to rename it to something different because "unsafe" sounds a lot scarier that what `routefUnsafe` is about and it has a lot of other meaning in .NET which could cause confusing for Giraffe users.

Krzysztof-Cieslak:
I’m building router dynamically and the only way to abstract over different instances of `PrintFormat` is to use `obj` as a result type and unbox result on runtime.

Happy to change the name, open for suggestions.

Krzysztof-Cieslak:
Ping ;)

gerardtoconnor:
Hi, can you give some code examples of what you are trying to do so we can figure out safe performant alternative as working with obj should be last resort given the type safety sacrificed

dustinmoris:
Hey, if you still need this functionality, could you please [open a new issue here](https://github.com/giraffe-fsharp/Giraffe.TokenRouter) now?

Issue #196, opened 2017-12-26
Readme should use requesterror module instead of setstatuscode 


forki:
In the readme we have many places where we use fish + setstatuscode. We should change that to simpler samples that use Requesterror module

dustinmoris:
Thanks @forki  for the suggestion and @vtquan  for the PR!

PR #209, opened 2018-01-18
Replace setStatusCode in Readme to StatusCode HttpHandlers.

vtquan:
Agreeing with #196, I changed the Readme to use the handlers instead of setStatusCode except when being used to contrast against the HttpHandlers. This would simplify the examples. I also left setStatusCode alone for 401 since RequestErrors.UNAUTHORIZED requires stating the scheme and realm.


dustinmoris:
Fixes #196

PR #211, opened 2018-01-19
Fixes for handling nullables when binding query strings

colinbull:
Previously I was getting an error when binding records with `Nullable<_>` properties.


colinbull:
Whoops! Forgot to pull latest develop.. I'll create a new one.

dustinmoris:
Sorry yes, I am just working on a new version but unfortunately the AppVeyor build still fails because they are still on an older SDK version. Looks like your fix is fairly small so hopefully my changes won't make it too difficult to apply to the latest dev again! Thank you!

PR #213, opened 2018-01-19
Fix example for RequestErrors.UNAUTHORIZED

vtquan:
RequestErrors.UNAUTHORIZED requires scheme and realm which were not given in the example. Trying to run the example would give a ``The type 'HttpFunc' does not match the type 'string'`` error.


dustinmoris:
Thanks for the fix!

Issue #215, opened 2018-01-26
Nested task ce issue with htmlFile when hosted on IIS


lambdakris:
## Description
I am having one of those seemingly async related issues with an app I deployed to an IIS server.

It is simply the use of the htmlFile handler, something like so...
```
route "/example" >=> htmlFile "wwwroot/example.html"
```
It runs fine when debugging or when launching from dotnet cli, but when I deploy it to an IIS server I begin to get the following exception:
```
System.UnauthorizedAccessException: Access to the path 'C:\inetpub\wwwroot\Example\wwwroot\example.html' is denied.
``` 
I have the wwwroot setup so that it is available both when running locally and when deployed and I have made sure it is not a path issue as you will see if you look at my analysis below

## The short analysis
I believe the issue stems from the interaction between the task computation(s) used within the htmlFile handler and how the FileStream object is initialized. Specifically, [line 22 in Common.fs](https://github.com/giraffe-fsharp/Giraffe/blob/master/src/Giraffe/Common.fs#L22) is insufficient for a situation where it is used from a nested task ce...probably because the FileShare mode is not set to read yet it gets read from the containing task while it has not yet been release from the nested task...?

## The long analysis
In case it provides anyone with something important, this is what I looked into.

Taking a guess that the issue had to do with async, I wrote a couple of variations of the logic within the [htmlFile handler](https://github.com/giraffe-fsharp/Giraffe/blob/master/src/Giraffe/HttpHandlers.fs#L341) and the [readFileAsStringAsync function](https://github.com/giraffe-fsharp/Giraffe/blob/master/src/Giraffe/Common.fs#L20) from scratch so that I could isolate the effects.

The following works (notice that everything happens inside a single task ce)
```
...
task {
    use stream = new FileStream(filePath, FileMode.Open)
    use reader = new StreamReader(stream)
    let! content = reader.ReadToEndAsync()
    return! html content next context
}
...
```
The following DOES NOT work (notice that the io ops are in a nested task ce)
```
...
task {
    let! content = task {
        use stream = new FileStream(filePath, FileMode.Open)
        use reader = new StreamReader(stream)
        return! reader.ReadToEndAsync()
    }
    return! html content next context
}
...
```
The following DOES work (notice that the io ops are still in a nested task ce but there is no explicit FileStream)
```
...
task {
    let! content = task {
        use reader = new StreamReader(filePath)
        return! reader.ReadToEndAsync()
    }
    return! html content next context
}
...
```
[...]

PR #216, opened 2018-01-27
Use StreamReader(String) in readFileAsStringAsync

lambdakris:
Use StreamReader(String) in readFileAsStringAsync, eliminating the need to explicitly initialize a FileStream

- addresses https://github.com/giraffe-fsharp/Giraffe/issues/215


dustinmoris:
Thanks for the PR. Would you mind making this PR against the `develop` branch please? I have changed the base but now there is a conflict and you're probably quicker in making the PR against develop than manually resolving the conflict now :)

Issue #224, opened 2018-02-10
Policy-based auth


Nhowka:
Looking at the asp.net docs, I found that feature could be useful. https://docs.microsoft.com/en-us/aspnet/core/security/authorization/policies

Is there a way to use it from Giraffe like the Roles?

dustinmoris:
Hi, policy based auth is on my list. I am aiming to have this ready for the 1.2.0 release. 1.1.0 is aimed at going out latest by the end of this week just to give you some time frame.

Nhowka:
After adding `Microsoft.AspNetCore.Authorization` to the project, I could add the authorization services using the code from [this article](https://vtquan.github.io/fsharp/setting-up-services-with-giraffe/). Then I created the method requiresPolicy as the following:

```F#
let requiresPolicy (policy : string) (authFailedHandler : HttpHandler) : HttpHandler =
    fun (next : HttpFunc) (ctx : HttpContext) ->                
        match ctx.GetService<IAuthorizationService>() with
        | null ->                     
            authFailedHandler finish ctx
        | authService -> task {            
            let! authResult = authService.AuthorizeAsync (ctx.User, policy)
            if authResult.Succeeded then
                return! next ctx
            else
                return! authFailedHandler finish ctx
            }
```

I tried to change the tests to use it, but I figured after a long time that it was not using the real deal and the errors were because the service was not injected. It worked in a different project, though.

dustinmoris:
This has been released with [Giraffe 2.0.0](https://github.com/giraffe-fsharp/Giraffe/releases/tag/v2.0.0).

PR #237, opened 2018-02-27
WIP: Policy-based authorization

Nhowka:
Address #224. It's working in another project, but the code could be polished further.

I'll need help with creating tests for it because of the dependency injection of `IAuthorizationService`.

### DIFF
--- a/src/Giraffe/Auth.fs
+++ b/src/Giraffe/Auth.fs
@@ -47,18 +47,16 @@ let signOut (authScheme : string) : HttpHandler =
 /// ** Output **
 /// A Giraffe `HttpHandler` function which can be composed into a bigger web application.
 let requiresPolicy (policy : string) (authFailedHandler : HttpHandler) : HttpHandler =
-    fun (next : HttpFunc) (ctx : HttpContext) ->                
-        match ctx.GetService<IAuthorizationService>() with
-        |null ->                     
-            authFailedHandler finish ctx
-        |authService -> task {            
+    fun (next : HttpFunc) (ctx : HttpContext) ->
+        let authService = ctx.GetService<IAuthorizationService>()
+        task {
             let! authResult = authService.AuthorizeAsync (ctx.User, policy)
             if authResult.Succeeded then
                 return! next ctx
             else
                 return! authFailedHandler finish ctx
             }
-        
+
 /// ** Description **
 /// Validates if a `ClaimsPrincipal` satisfies a certain condition. If the `policy` returns `true` then it will continue with the `next` function otherwise it will shortcircuit to the `authFailedHandler`.
 ///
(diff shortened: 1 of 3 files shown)


src/Giraffe/Auth.fs
```diff
@@ -36,6 +37,28 @@ let signOut (authScheme : string) : HttpHandler =
...
+///     - `policy`: A `string` containing the desired policy to validate
+///     - `authFailedHandler`: A `HttpHandler` function which will be executed when the policy is not satisfied.
+///
+/// ** Output **
+/// A Giraffe `HttpHandler` function which can be composed into a bigger web application.
+let requiresPolicy (policy : string) (authFailedHandler : HttpHandler) : HttpHandler =
+    fun (next : HttpFunc) (ctx : HttpContext) ->                
+        match ctx.GetService<IAuthorizationService>() with
+        |null ->                     
+            authFailedHandler finish ctx
```
dustinmoris: Not sure if we should invoke the `authFailedHandler` when the `IAuthorizationService` hasn't been registered. It might give the user the wrong impression of where the error lies. I think I'd prefer to just let a NRE bubble up, or change the `ctx.GetService<'T>` implementation to check for `null` and throw a meaningful exception in one place for all `GetService` calls. What do you think?

dustinmoris:
Thanks for this PR. I was actually having something similar already in the works but even better if someone else has already done it :).

Questions:

- Do we need to add `services.AddAuthorization()` as part of the default Giraffe dependencies when calling `services.AddGiraffe()` (probably not because it wouldn't make sense to register without any policies, but just checking)?
- `requiresPolicy "nameOfPolicy"` is good, because it matches MVC, but would it make sense to add another handler which doesn't rely on a magic string? e.g. `requiresPolicy<MyPolicy>`?

Would be also good to add an example of it to one of the existing sample projects and some unit tests.

src/Giraffe/Auth.fs
Nhowka: Probably use the `GetRequiredService` would be fitting as it already throws on missing service, but yeah, agree with the first part.

Nhowka:
For the former, it would do no harm and would make the null check redundant as it would never match. I thought about adding, but it wasn't essential.

As for the latter, I don't know how to start. Looks like every policy gets registered with a string, so I'm not sure if it can be made in a less magic way unless we change the way policies get registered.

I'll try to add something using it to the samples, but I'll need help with the unit tests. I couldn't inject the services into context when I tried to write something.

src/Giraffe/Auth.fs
Nhowka: Looks like we can't use the `GetRequiredService` on that context, so it could be nice as a method extension to throw a fitting exception.

Nhowka: Didn't see that `GetService` was already an extension. Should be ok to throw on missing. I'll add something to a future commit.

dustinmoris:
Hey, thanks for the additional changes. I pulled your branch and started adding samples and tests. I'll push to your branch tonight and then I'll merge this PR!

Issue #238, opened 2018-03-01
Route format with System.Guid placeholders don't work


ebi-igweze:
When using the route formatter i.e. "**routef** somePath" with guid placeholders i.e. **%O**, if you have multiple placeholders only the first placeholder gets transformed by the route formatter, and on another occasion, it formats the first one as a duplicate.
````
i.e
let myHandler: HttpHandler = 
     fun ( parentId, childId) -> text (sprintf "%O is the parent to: %O" parentId childId)

let webApp = 
    choose [
          GET routef "/api/families/%O/children/%O" myHandler
         statusCode 404 >=> text "Not found"
     ] 
````

**Case 1:** if you tried a GET request using "api/families/4ec87f064d1e41b49342ab1aead1f99d/children/2a6c918595d94d8c80a6575f99c2a716"
only the first Guid gets evaluated and passed into the function 'myHttpHandler', causing an error at run-time.

**Case2:** Now if you try the other Guid format with the hyphens i.e. "api/families/4ec87f06-4d1e-41b4-9342-ab1aead1f99d/children/2a6c9185-95d9-4d8c-80a6-575f99c2a716" the first Guid (i.e. 4ec87f06***) is what gets evaluated as both the parentId and childId.

Please let me know if there is something I am missing. because I have been stuck with this problem for a while now.

**Comment by hxdhttk** [comment]
I think change the regex for Guid would fix this.
```fsharp
let private formatStringMap =
    let guidFormatStr =
        "(([0-9A-Fa-f]{8}\-[0-9A-Fa-f]{4}\-[0-9A-Fa-f]{4}\-[0-9A-Fa-f]{4}\-[0-9A-Fa-f]{12})|([0-9A-Fa-f]{32}))"
```
The regex should be fixed as `"([0-9A-Fa-f]{8}\-[0-9A-Fa-f]{4}\-[0-9A-Fa-f]{4}\-[0-9A-Fa-f]{4}\-[0-9A-Fa-f]{12}|[0-9A-Fa-f]{32})"` to avoid capturing more groups that are not represented by the route path.
[link to the code](https://github.com/giraffe-fsharp/Giraffe/blob/4b5852e80e49c11e11bb66a4065afd79b6ad778b/src/Giraffe/FormatExpressions.fs#L15)

ebi-igweze:
Currently, the workaround I had to do was represent the Guid using string placeHolders and converting the string representation to Guid within the HttpHandlers. i.e.
```F#
[...]

PR #239, opened 2018-03-05
Fix regex for capturing Guid to make routef work for route paths which have multiple Guid placeholders.

hxdhttk:
This PR may fix the issue addressed in [#238](https://github.com/giraffe-fsharp/Giraffe/issues/238).
routef now failed to capture all Guids in a route where there are multiple Guid placeholders (`%O`).


dustinmoris:
Awesome, thanks for the fix! Would you mind adding one unit test to check for the case of multiple Guids so that we have a breaking test if this regex changes again?

hxdhttk:
Just added a test in `RoutingTests.fs`.

dustinmoris:
This PR is part of the 1.2.0-preview-1 release for now.

Issue #255, opened 2018-04-11
GiraffeViewEngine - button onclick event is missing


TheDevKnight:
Hi,

is the onclick event on the button really not available?

button [ _onclick ??? ] [ ]

As a workaround i could add the event in my javascript. Is this the correct way?

Thanks for help.

Regards

dustinmoris:
Hi, if the _onclick event is missing then you can [create it yoursef](https://github.com/giraffe-fsharp/Giraffe/blob/master/DOCUMENTATION.md#custom-elements-and-attributes) or submit a PR!

TheDevKnight:
@dustinmoris Thank you! Works perfect and very easy :)

dustinmoris:
This has been released with [Giraffe 2.0.0](https://github.com/giraffe-fsharp/Giraffe/releases/tag/v2.0.0).

PR #256, opened 2018-04-12
Added attribute for onclick event

TheDevKnight:
#255


TheDevKnight:
@danieljsummers thx for the hint, i change it

Issue #254, opened 2018-04-10
URLDecoding + sign is missing


rmotyka:
When I do manually:
```
let temp = WebUtility.UrlDecode("Work%20Instruction%20Manual%20D-
touch%20Rev.%201%20Eng%2BPL.pdf")
```

it gives me:  Work Instruction Manual D-touch Rev. 1 Eng+PL.pdf

but using:

`routeCif "/download/%s"  (fun x -> handleDownloadDocument x)
`
x is: Work Instruction Manual D-touch Rev. 1 Eng PL.pdf (space instead of +)

I've downloaded master version of Giraffe and made a test:

```
[<Fact>]
let ``String with plus`` () =
    tryMatchInput "/foo/%s" "/foo/Work%20Instruction%20Manual%20D-touch%20Rev.%201%20Eng%2BPL.pdf" false
    |> function
        | None -> assertFail "Format failed to match input."
        | Some (s1 : string) ->
            Assert.Equal("Work Instruction Manual D-touch Rev. 1 Eng+PL.pdf", s1)
```
but it works fine. 

Is it possible if the UrlDecode is done several times over the string? If so, the last run of UrlDecode removes the plus sign...

It seems that _ctx.Request.Path.Value_ contains partially decoded string. All is decoded except %2F ie "/" sign. It makes sense, as that might interfere with path signs:
/download/filepath
where filepath = %2Fhome%2Froman%2FDocuments (one parameter). If we decode it completely:
/download/home/roman/Documents - not what we wanted.

dustinmoris:
Hmm interesting...

I think it would be better to write a test which runs through the entire pipeline (like [this](https://github.com/giraffe-fsharp/Giraffe/blob/master/samples/SampleApp/SampleApp.Tests/Tests.fs)).

Could you perhaps provide a test like this which shows the error and I'll have a look what needs to be done to fix it! Thanks for reporting.

dustinmoris:
This has been released with [Giraffe 2.0.0](https://github.com/giraffe-fsharp/Giraffe/releases/tag/v2.0.0).

PR #257, opened 2018-04-12
SampleApp showing not proper url decoded string. 

rmotyka:
a%2Fb%2Bc.d%2Ce 
gives 
a/b c.d,e 
and should gives 
a/b+c.d,e #254


rmotyka:
It seems that string is already decoded, only what left is "/" char.

Issue #264, opened 2018-05-07
BUG: HTTP Error 502.5 - Process Failure w/ Giraffe Template: `dotnet publish` & `dotnet run`


wiredsister:
Steps to reproduce: 
1.  Install template with: `dotnet new giraffe -lang F#`
2. Build and publish with `dotnet publish` (also tried: `dotnet publish -c Release -r win10-x64`)
3. Do `dotnet run`
4. When you open your browser to the hosted port (by default: `localhost:1221` you either see nothing or you see the exn below-- I had to click `View` in VS2017 to see this at all.)

```
HTTP Error 502.5 - Process Failure

Common causes of this issue:
The application process failed to start
The application process started but then stopped
The application process started but failed to listen on the configured port

Troubleshooting steps:
Check the system event log for error messages
Enable logging the application process’ stdout messages
Attach a debugger to the application process and inspect

For more information visit: https://go.microsoft.com/fwlink/?LinkID=808681
```

I have enabled log files and they are just empty but being created each time I run the app. I have tried specifying the full path to my Giraffe dll inside `arguments` property on the `aspNetCore` tag in my `web.config`. I also tried using the full path to my .exe as the `processPath` property on that `aspNetCore` tag. Same situation. 

I also tried troubleshooting the tips here: https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/azure-iis-errors-reference?view=aspnetcore-2.1 but found nothing useful. 

Let me know if I can provide any more information! I'll reference this ticket if I can fix and PR into the `giraffe-template` project. I'm working with Enrico Sada (@enricosada on Github) and he was able to reproduce this bug with me.

Gina

PS: 

```
> dotnet --info
.NET Command Line Tools (2.1.104)

Product Information:
 Version:            2.1.104
 Commit SHA-1 hash:  48ec687460

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.10586
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\2.1.104\

Microsoft .NET Core Shared Framework Host

  Version  : 2.0.6
  Build    : 74b1c703813c8910df5b96f304b0f2b78cdf194d
```

&  using `Giraffe (1.1)`

cartermp:
@wiredsister I don't have an answer, but what happens if you do `dotnet path-to-dll`, where `path-to-dll` is the path to the win10 binary in the publish output?

[...]

PR #265, opened 2018-05-07
Update README instructions for Visual Studio 2017 to include correct IIS information.

wiredsister:
See https://github.com/giraffe-fsharp/Giraffe/issues/264 for details.


dustinmoris:
Hey, so after more consideration I think I have to close this PR without a merge. The giraffe project is just a class library and not a web project, so in the context of this README the note on IIS doesn't make much sense and also even if I wanted to add this modification to the README of the giraffe-template I don't think it is true that giraffe will fail if using IIS Express. The point is that any ASP.NET Core application will not work in IIS without a web.config and as such there is nothing unique about Giraffe.

Hope my explanation makes sense, but if you disagree then please feel free to open a new issue.

PR #273, opened 2018-06-02
Fix one typo in README.md

0x53A:
tempalte => template


dustinmoris:
Actually I just see that this typo has already been fixed through another PR which has been merged into develop before, but appreciate taking the time to bring this to light again!

PR #285, opened 2018-07-10
Fix Guid Regex

mrinaldi:
The previous regex creates three match groups instead of just one. That
was returning a wrong value when there was another placeholder after
the %O.


mrinaldi:
I just noticed this is on 1.2.0-preview-1.
Closing PR.

PR #300, opened 2018-09-01
Statefull rendering of XmlNode

dv00d00:
- Added several questionable optimizations for EncodedText and attr (do html encoding during before rendering). It does not affect current impl much (2ms / 200 bytes win for this particular benchmark) but doubles the perf of statefull implementatoin
- Implemented xml and html rendering of XmlNode to String Builder
- Added benchmarks for html rendering
- Benchmark app got its very own StringBuilder cache to investigate the impact of having it. Would like to discuss it further

Open questions:

- Naming and code style. Please let me know if I need to change something
- HttpHandlers to actually use StringBuilder based version
- Could we afford to have #if directives to utilize span goodies from core 2.1?

Becnhamrk results from my pc:

```
BenchmarkDotNet=v0.11.0, OS=Windows 10.0.17134.228 (1803/April2018Update/Redstone4)
Intel Core i5-3570K CPU 3.40GHz (Ivy Bridge), 1 CPU, 4 logical and 4 physical cores
Frequency=3420544 Hz, Resolution=292.3512 ns, Timer=TSC
.NET Core SDK=2.1.400
  [Host]     : .NET Core 2.1.2 (CoreCLR 4.6.26628.05, CoreFX 4.6.26629.01), 64bit RyuJIT DEBUG
  DefaultJob : .NET Core 2.1.2 (CoreCLR 4.6.26628.05, CoreFX 4.6.26629.01), 64bit RyuJIT
```

|                    Method |      Mean |     Error |    StdDev | Scaled |   Gen 0 | Allocated |
|-------------------------- |----------:|----------:|----------:|-------:|--------:|----------:|
|        RenderHtmlOriginal | 42.010 us | 0.3062 us | 0.2714 us |   1.00 | 16.2354 |   49.9 KB |
|       RenderHtmlStatefull |  2.246 us | 0.0127 us | 0.0118 us |   0.05 |  1.4381 |   4.42 KB |
| RenderHtmlStatefullCached |  1.867 us | 0.0146 us | 0.0129 us |   0.04 |  0.6275 |   1.93 KB |


dustinmoris:
Wow awesome, I had a brief look and this looks nice. I will spend some time on this PR tomorrow and get back to the points you mentioned as I am away from my desk today, but just wanted to say thanks for the effort.

dustinmoris:
Ok, so here are some thoughts and questions:

- It seems to me that the `StringBuilder` version is doing exactly the same as the current version, except more efficient, correct? Why did you name it Stateful? I didn't see anything stateful about the `StringBuilder` implementation? (I am asking, because ideally I would like to replace the implementation of the current with your implementation and then we don't need to think about naming either)
- "Could we afford to have #if directives to utilize span goodies from core 2.1?"
I'd say yes. Given that netstandard2.1 will have spans as well I think sooner or later we will have to introduce #if directives anyway, because I don't think Giraffe will stop support for netstandard2.0 as soon as 2.1 comes out.

slang25:
You are spot on with point 2, I think that is a good strategy and imagine that many other libraries will be doing the same 6-12 months from now.

dv00d00:
Stateful in a sense that it does return unit performing mutations on a passed string builder instance. But yes, it is true that we can hide the implementation details under the hood. There is still value in having the pure implementation, at least to have it as a reference, because sb stuff is not so easy to follow up.

And then the string builder instance management comes into play, We could new up a fresh string builder every request. Or we can consume a thread static instance. Thread static is nice in a way that you will get a singleton per thread, but it also means you can't just pass sb instances around, especially in continuations.

I'd still go with string builder cache, but not expose it as a public api. 

Will push updated handlers soon, we'll see how it goes.

dustinmoris:
Ok, so I see what you are doing here with the cache. You basically try to avoid to new up a StringBuilder as often as possible, by reusing a StringBuilder after its usage.

From your benchmark tests I can see that not always creating a new StringBuilder instance has given you a small edge over the test which always creates a new instance, however I think your current implementation is not thread safe.

Looking at this code:

```fsharp
    static member Get(capacity:int) = 
        
        if capacity <= MaxBuilderSize then
            let ms = MemoryStreamCache.instance;
            let capacity = max capacity DefaultCapacity
            
            if ms <> null && capacity <= ms.Capacity then
                MemoryStreamCache.instance <- null;
                ms.Clear()
            else
                new StringBuilder(capacity)
        else
            new StringBuilder(capacity)
```

It is entirely possible that you have two simultaneous requests/threads coming in which get a non null `StringBuilder` instance from `let ms = MemoryStreamCache.instance` and then therefore both pass the check of ` if ms <> null && capacity <= ms.Capacity then`, which will lead to the undesired consequence that you are going to mutate one StringBuilder by two requests at the same time.

In order to make this thread safe you'd have to [introduce (double-) locking](http://csharpindepth.com/Articles/General/Singleton.aspx) at least, which I believe will revert the minor perf gain of not having to new up a `StringBuilder` each time.

This is my feeling by looking at the current implementation, but I might be missing something and be entirely wrong.

Maybe if you could explain me how you thought this would work?

slang25:
Because `MemoryStreamCache.instance` is marked as `ThreadStatic`, the compiler ensures that there is only one instance per thread, which means you can revert to non-thread safe access.

dustinmoris:
Ok nevermind, ignore my previous message, I didn't see that you decorated the instance with `[<ThreadStatic>]`. Now I get it.

dustinmoris:
Ok, so now that I think I've got the full picture here are my thoughts:

- I think we should replace the current view engine rendering functions with the ones which utlise a StringBuilder, because I think there's no point in having two sets if both deliver the same result, but one is significantly faster than the other.

- I consider the StringBuilder cache a micro optimisation which adds reasonable complexity for a rather minor gain. Having said that I also think that the added complexity is rather small in actual code and if we can encapsulate it in one place and hide it from the ViewEngine logic then I'd be happy to add that complexity for that extra perf gain.

My first instinct tells me to create a factory for the StringBuilder which can be called from the ViewEngine and can also be overridden by users if they want to go with a simpler model. Kind of similar to what Giraffe currently does with [serializers](https://github.com/giraffe-fsharp/Giraffe/blob/master/DOCUMENTATION.md#json).

What are your thoughts on this?

dv00d00:
Ok, I've done latest changes before I was reading the comments, I'll try to hide string builder cache behind some interface and remove it from the public API surface of rendering methods.

dv00d00:
I've tried to add a new interface for view rendering, it felt too involved.

I've removed StatfullRendering namespace
Added ticked versions of render functions
Preserved the shape of existing rendering functions
Made StringBuilder cache private
HtmlViewWriter uses string builder caching and pooled array to copy chars for conversion to utf8

dustinmoris:
How come you removed the `static member Get(capacity:int)` method from the cache? Now the `StringBuilder` might not be large enough if a view is larger than `DefaultCapacity`, right?

Nhowka:
That may be a dumb question, but wouldn't `bprintf` be a good fit for what is being achieved? It takes a `StringBuilder` and can write formatted strings to it, as seem [here](https://msdn.microsoft.com/en-us/visualfsharpdocs/conceptual/printf.bprintf%5B't%5D-function-%5Bfsharp%5D).

dv00d00:
@dustinmoris So previously I was never calling a StringBuilderCache.Get with anything else than DefaultCapacity. The lifetime of the cached stringbuilder instance with this code is:

1. No instance cached in the thread yet, return new with default capacity
2. A used instance is released back to the cache. If it has grown to much ( > MaxCapacity), Release method won't accept it and GC will eventually collect it, otherwise, the cache will store it and reuse for subsequential requests.

Btw Currenlty max capacity is set to 64k of chars which is 128kb or RAM, meaning we will potentially hold a ref to the LOH object. Which brings another set of considerations...

Tbh it took me a good 10 minutes to read those 20 lines of code in sb cache, I don't have strong feelings about it. Maybe it would be better to let the end user decide how to manage StringBuilders lifetime. 


@Nhowka I've tested format method on the StringBuilder itself, they are performing much worse than their single param siblings. I will check it in bench though, thx for the hint. Did not know such overloads existed, sigh.

dv00d00:
I've wrapped StringBuilder cache under new interface in Serialization module, looks much better now imo.

dv00d00:
Did a `bprintf` test

```fsharp
[<MemoryDiagnoser>]
[<ShortRunJob>]
type FormattersBenchmark() = 
    
    let tag = "div"
    let attr = "class"
    let attrV = ".test"

    [<Benchmark(Baseline = true)>]
    member this.Manual() =
        let sb = StringBuilderCache.Get()
        sb += "<" += tag += " " += attr += "=\"" += attrV += "\"" +! "/>"
        StringBuilderCache.Release sb

    [<Benchmark>]
    member this.Auto() = 
        let sb = StringBuilderCache.Get()
        Printf.bprintf sb "<%s %s=\"%s\" />" tag attr attrV
        StringBuilderCache.Release sb
```

| Method |      Mean |     Error |   StdDev | Scaled | ScaledSD |  Gen 0 | Allocated |
|------- |----------:|----------:|---------:|-------:|---------:|-------:|----------:|
| Manual |  88.38 ns |  31.03 ns | 1.753 ns |   1.00 |     0.00 |      - |       0 B |
|   Auto | 807.48 ns | 161.60 ns | 9.131 ns |   9.14 |     0.17 | 0.1822 |     288 B |

Nhowka:
Yikes! Thanks for testing it!

dustinmoris:
Hi, I just pushed a few changes into this PR with some ideas up for discussion. I will explain my changes a little bit later. Need to rush to work now :)

dustinmoris:
Some of the build script changes and other unrelated stuff can be ignored from my last commit. These changes appear in this PR, because the PR was opened against master (from a master fork), but I ask all PRs to go into develop first please, so I pulled the latest develop into this PR to make sure there are no conflicts and then I made a few minor changes to the actual code related to this PR. As said in my previous comment, I'll explain what I've done so we can discuss it a bit later!

dv00d00:
Several observations:

- Breaking change in XmlNode DU, will this mean a new major version? I have several not published proposals on JsonSerializer on the way which would benefit from breaking changes 😄 
- Dispose on a factory perf is the same as in manual release approach. But it looks a bit alien because the first impression is that factory instance would be disposed not its products.
- We should probably have an invariant on `ThreadStaticStringBuilderCache` which demands defaultCapacity < maxCapacity
- It looks like in the current version, StringBuilder instance capacity is not considered when asking for a cached instance


```fsharp
    static member Get (capacity : int) (maxCapacity : int) : StringBuilder =
        match StringBuilderCache.inUse with
        | true  -> new StringBuilder(capacity)
        | false ->
            StringBuilderCache.inUse <- true
             let sb = StringBuilderCache.sb
             match sb <> null && sb.Capacity >= capacity with //always true
            | true  -> sb.Clear()
            | false ->
                let sb' = new StringBuilder(capacity)
                if capacity <= maxCapacity then // always true
                    StringBuilderCache.sb <- sb'
                sb'
```
Maybe something like that could work
```fsharp
    static member Get (defaultCapacity : int) (maxCapacity : int) : StringBuilder =
        match StringBuilderCache.inUse with
        | true  -> new StringBuilder(defaultCapacity)
        | false ->
            StringBuilderCache.inUse <- true

            match StringBuilderCache.sb with
            | sb when sb <> null && sb.Capacity < maxCapacity -> 
                sb.Clear()
            | _ -> 
                StringBuilderCache.sb <- new StringBuilder(defaultCapacity)
                StringBuilderCache.sb
```

dustinmoris:
Ok so here some thoughts on my last commit:

- There's two sets of view functions: `build[...]` and `render[...]`. The first is a new set of functions which accepts and returns a `StringBuilder` instance with the view (e.g. `buildHtmlDocument`). The latter just returns a single `string` with the entire rendered view (like before, e.g. `renderHtmlDocument`).
- Replaced the `RawText` and `EncodedText` union types with a single `Text`, since we moved the encoding of text into the helper functions.
- Renamed the interface to `IStringBuilderProvider` and replaced the `Release` function with the `IDisposable` interface. I am not 100% sure this is the best way, but I think that the interface should not be tightly coupled to the implementation of a single concrete type. The `Release` method is really only an implementation detail of this particular cached version of the `StringBuilder` provider and therefore wouldn't make sense for any other implementation. An `IDisposable` feels cleaner IMHO, because it offers a more standardised way of doing any cleanup/disposal work which could make sense for other implementations as well.

However, overall I am not sure if this isn't an overkill now. Realistically we are introducing a new interface now and types which do nothing but giving me a new instance of a `StringBuilder`. I don't see how any user might find this useful beyond the specifics of our view engine. We don't have interfaces and types for other objects wither (e.g. `MemoryStreamProvider` to issue a cached memory stream, `ByteArrayProvider` to reuse a byte array, etc.).

After more thought I think we are overcomplicating this. I think we should make a hard decision now and implement the view engine in one way and hide everything else from the user, because I don't want to bother users to have to think about how to provide a `StringBuilder`. Most people will be saying why shall I even think about it, it's just a `StringBuilder`, just create a new one...

So, I think we should:

- Make the `build[...]` functions private, so users don't have to think about whether to use `renderHtmlDocument` or `buildHtmlDocument`. They get to use just the `render[...]` one.
- Make the `render[...]` functions be responsible for deciding if they use the cached or non cached string builders. The only difference is perf, so either everyone should benefit from it (and therefore the function should just use the one) or nobody.

I'd be happy to go with the cached version. What do you think?

dv00d00:
From the top of my head. I'd remove caching stuff altogether and leave StringBuilder rendering functions public but put them in a module with qualified access. So the user or 3rd party package could build on top of that.

I agree that there is a little sense in exposing new interface just for the view engine. Maybe the proper abstraction will arise later.

dustinmoris:
Based on your last comment I removed the interface again. @dv00d00 please review this PR and submit any changes you see fit. Let's get this done and dusted so we can release this nice perf improvement soon! :)

dv00d00:
Fixed benchmarks, I think it is ready to go. Thanks for the cleanup @dustinmoris !

dustinmoris:
Thank you @dv00d00 !!!

Issue #301, opened 2018-09-10
Proposal: Extend IJsonSerializer with UTF8 serialization methods


dv00d00:
There are several json serializers which supports serialization to utf8 directly without the need to allocate intermidiate string which the would be converted to utf8. 

Several examples :

- https://github.com/neuecc/Utf8Json
- https://github.com/Tornhoof/SpanJson

I propose to extend IJsonSerializer  with new methods which will allow to utilize utf8 serialization features.

```fsharp
    [<AllowNullLiteral>]
    type IJsonSerializer =
        abstract member Serialize<'T> : 'T -> string // generic on input
        abstract member SerializeToUtf8<'T> : 'T -> byte[] // produce output as a utf8 byte array
        abstract member SerializeAsync<'T> : 'T * Stream -> Task // write utf8 to stream

        abstract member Deserialize<'T> : string -> 'T
        abstract member Deserialize<'T> : Stream -> 'T
        abstract member DeserializeAsync<'T> : Stream -> Task<'T>
```

Generic argument on serialize will allow to pass structs without heap allocations. 

Sample implementation using Utf8Json

```fsharp
    type Serializer(resolver: IJsonFormatterResolver) = 
        interface IJsonSerializer with

            member __.Serialize<'T> (o : 'T) = 
                JsonSerializer.ToJsonString o

            member __.SerializeToBytes<'T>(o : 'T) = 
                JsonSerializer.Serialize(o, resolver)

            member __.SerializeAsync<'T>(o : 'T, dest: Stream) = 
                JsonSerializer.SerializeAsync(dest, o, resolver)

            member __.Deserialize<'T>(json : string) = 
                JsonSerializer.Deserialize<'T>(json, resolver)

            member __.Deserialize<'T>(stream : Stream) = 
                JsonSerializer.Deserialize<'T>(stream, resolver)

            member __.DeserializeAsync<'T>(stream : Stream) = 
                JsonSerializer.DeserializeAsync<'T>(stream, resolver)
```

dustinmoris:
Sounds good to me! Utf8Json looks great!

dustinmoris:
Release is out: https://github.com/giraffe-fsharp/Giraffe/releases/tag/v3.0.0

Thanks for the great contributions!

PR #303, opened 2018-09-13
Json serializer enhancements

dv00d00:
As discussed in #301 I am adding

- a generic parameter for serialization to string to avoid value types boxing. (Json serializer could have the fast path for value types)
- serialization to a utf8 byte array
- async serialization to utf8 stream

Open questions:

- new repos for utf8 serializers
- chunked transfer json response writer in the core repo

### DIFF
--- a/src/Giraffe/Serialization.fs
+++ b/src/Giraffe/Serialization.fs
@@ -10,7 +10,7 @@ module Json =
     open System.Threading.Tasks
     open Newtonsoft.Json
     open Newtonsoft.Json.Serialization
-    open FSharp.Control.Tasks.V2.ContextInsensitive
+    open System.Text
 
     /// **Description**
     ///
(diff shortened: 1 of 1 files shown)


dustinmoris:
Ok I have made the following additional changes:

- Renamed all `Serialize...` methods to the output type they serialize to, because `Serialize` on its own would have caused naming conflicts and if only one method was named `SerializeToBytes` then it felt inconsistent. Now there is `SerializeToString`, `SerializeToBytes` and `SerializeToStreamAsync`.
- Changed `SerializeToStreamAsync<'T> : 'T * Stream -> Task` to `SerializeToStreamAsync<'T> : 'T -> Stream -> Task`
- Added new `Utf8JsonSerializer` and swapped it in as the new default JSON serializer in the `AddGiraffe()` method which registers all dependencies
- Added `Transfer-Encoding: chunked` HTTP header to the two new JSON HTTP handler functions.

I am just playing and experimenting but I think swapping Newtonsoft.Json with Utf8Json would make sense. What do you think?

Also do you think Giraffe should by default implement [Utf8Json.FSharpExtensions](https://github.com/pocketberserker/Utf8Json.FSharpExtensions)?

dv00d00:
Newtonsoft could be a little bit slower and not provide the proper async interface to serialization methods, but it is battle tested in 100M downloads. Utf8Json currently only has 80k or less, could break on F# specific types such as list or option. 

Utf8Json.FSharpExtensions is probably should be used to smooth adoption a bit. Would be nice to hear from @pocketberserker.

`Transfer-Encoding: chunked` is a nice touch, thx, previously it was implicit. 

I'd say to leave Newtonsoft as a default, for now, there might be some code depending on it in production.

dustinmoris:
> I'd say to leave Newtonsoft as a default, for now, there might be some code depending on it in production

Fair point.

So is there any other changes you would suggest or can this be merged?

dv00d00:
I've parametrized tests for json with newtonsoft / utf8 resolvers where appropriate

PR #304, opened 2018-09-14
enable SourceLink

ctaggart:
See https://github.com/ctaggart/SourceLink/issues/367 and the blogs for info on SourceLink in general: https://github.com/ctaggart/SourceLink/wiki/Blog-Posts about 

Please "Squash Merge" if you accept this pull request.

PR #305, opened 2018-09-14
Microsoft.NET.Test.Sdk does not need Program.fs anymore

ctaggart:
fixes https://github.com/giraffe-fsharp/Giraffe/pull/304#issuecomment-421491668


ctaggart:
There you go @dustinmoris. This will make `master` green again.

PR #307, opened 2018-09-22
Micro tuning of response writers

dv00d00:
This is one file micro tuning pr based on results of the continuous run of techempower [tests ](https://www.techempower.com/benchmarks/#section=test&runid=a1843d12-6091-4780-92a6-a747fab77cb1&hw=ph&test=plaintext&p=zik0zj-zik0zj-zijocf-5m9r)

image

There are 330_000 requests per second difference for this kind of benchmark between the utf8 encoding on each call vs utf8 encoding once in the constructor.

So I've moved utf8 encoding out of HTTP handlers where applicable.

I have not found a way to do the same thing for json because current serializer depends on an instance resolved from HTTP context


dustinmoris:
Awesome, that looks like smart fine tuning! Thanks!

PR #308, opened 2018-09-22
Added StringBuilderCache

dustinmoris:
Added string builder cache for review.
/cc @dv00d00


src/Giraffe/ResponseWriters.fs
```diff
@@ -9,16 +10,47 @@ open Microsoft.Net.Http.Headers
 open FSharp.Control.Tasks.V2.ContextInsensitive
 open Giraffe.GiraffeViewEngine
 
+
 // ---------------------------
 // Helper functions
 // ---------------------------
 
+let private DefaultCapacity = 1386
+let private MaxBuilderSize  = DefaultCapacity * 3
```
dustinmoris: @dv00d00 Would like to discuss these limits. How did you conclude `1386` as the default size and 3 times of it as the max size?

src/Giraffe/ResponseWriters.fs
```diff
@@ -9,16 +10,47 @@ open Microsoft.Net.Http.Headers
...
+
+type StringBuilderCache =
```
dustinmoris: Shall we hide this type entirely by making it private?

src/Giraffe/ResponseWriters.fs
```diff
@@ -9,16 +10,47 @@ open Microsoft.Net.Http.Headers
...
+
+    [<ThreadStatic>]
+    [<DefaultValue>]
+    static val mutable private instance : StringBuilder
+
+    static member Get () = StringBuilderCache.Get(DefaultCapacity)
+    static member Get (capacity : int) =
```
dustinmoris: If `nodeToUtf8HtmlDoc` always calls `Get()` do we even need the overload with `capacity : int` ?

src/Giraffe/ResponseWriters.fs
```diff
@@ -9,16 +10,47 @@ open Microsoft.Net.Http.Headers
...
+
+        if capacity <= MaxBuilderSize then
```
dustinmoris: If we always call `Get()` without a capacity (where the default capacity is used) then this if statement will always be true.

dustinmoris:
If we make use of the `StringBuilderCache` by default for the `nodeToUtf8HtmlDoc` function, then based on its current usage I think we could trim it down to:

```fsharp
let private DefaultCapacity = 1386

type private StringBuilderCache =
    [<DefaultValue; ThreadStatic>]
    static val mutable private instance : StringBuilder

    static member Get () =
        let sb = StringBuilderCache.instance
        if sb <> null then
            StringBuilderCache.instance <- null
            sb.Clear()
        else new StringBuilder(DefaultCapacity)

    static member Release (sb: StringBuilder) =
        if sb.Capacity <= (DefaultCapacity * 3) then
            StringBuilderCache.instance <- sb
```

dv00d00:
Random thoughts:

It will be difficult to pick a general purpose string builder cache size because it is coupled with the size of html being generated. In the techempower benchmark it is safe to state that the size would be 1024 or less because we know upfront the html size upfront. In other words it looks like a hack to get perf boost.

In order to make it less hacky we could introduce some sort of an end user controlled size, maybe via DI as we did previously in html PR. 

There is also a runtime threshold which we should consider. If an object is larger than 95k, it will be placed in the large object heap, where it would be pretty expensive to remove. We could get away by introducing invariant on the max size, but it also adds to overall complexity.

On the other hand, if the capacity of the string builder after all the manipulations would be larger than we want cache would behave like a regular new, so there still might be the benefit for a use case of small html pages

dv00d00:
The code in the benchmark is the direct port of C# code from techempower benchmarks repo, which probably was taken from the .net internals.

We could and probably should make it simpler and more F# friendly if we choose to follow this route.

I think we had pretty good version in the html pr.

dv00d00:
Also, not related to this pr, but techempower requires content type to include encoding for html bench. If we introduce an option to do so, stock implementation would be 4 lines shorter.

dustinmoris:
> In other words it looks like a hack to get perf boost

Agree, that's why I think we should consider to get the cache into Giraffe. I think we can think about different limits or maybe ways to let a user customize the default and max size.

> I think we had pretty good version in the html pr.

Cool, I'll see if I can dig it out again from history!

dustinmoris:
Ok, so thinking about it more I think configuring it is not really required.

The way I see it should work is:

- Keep the cache per thread initially reasonably small
- If a web app has larger views then the `StringBuilder` per thread will increasingly grow until it reaches the maximum size
- Keep the large `StringBuilder` instances in the cache if the app uses large `StringBuilder` objects a lot, otherwise start expiring the old big ones and replace with small ones again
- Set the maximum capacity to a fairly large character limit. If you think it shouldn't exceed 95kb then we can set this (~ 48000 characters)

dv00d00:
I think we can go with it, if end user won't be able to interact with except maybe configure it somehow, this should work nicely. 

Due to character limit (MaxCapacity) the implementation details of StringBuilder may, and probably will change in the future, the same is true for LOH size. So for now, I'd suggest to pick a safer bet and to allow of maximum 80000 bytes or 40000 chars.

Also, as it essentially will stick forever, we will allocate up to 40k ram per thread, which may, or may not cause issues, so there should be a way to disable it entirely.

dustinmoris:
I set the max limit to 40k like you suggested. The caching can also be disabled via `StringBuilderPool.AlwaysCreateNew <- true`. This property is `false` by default.

Unless there's any more ideas or objections I think this can be merged?

forki:
Doesn't the stringbuilder already have a cache inside the framework itself?

dustinmoris:
Not sure, what type of caching is `StringBuilder` doing?. Cache is maybe also the wrong wording here, because we are not really caching any content, we are only trying to avoid the repeat creation of a large object, similar to what `ArrayPool` is doing. That's why I renamed it to `StringBuilderPool` in my recent commits as well. There's definitely a perf improvement in the TechEmpower Benchmark tests to be seen if we make use of already existing StringBuilder objects though.

forki:
Yes I know. And that's exactly what they already do in the framework. IIRC
then e.g. String.Concat is using this thing under the hood. You may want to
take a look at the implementation.

Dustin Moris Gorski <notifications@github.com> schrieb am Do., 27. Sep.
2018, 09:52:

dustinmoris:
@forki: Ok thanks, will search for it and let you know!

dustinmoris:
I can't see any pooling of existing instances in the [StringBuilder](https://github.com/dotnet/corefx/blob/master/src/Common/src/CoreLib/System/Text/StringBuilder.cs) implementation. Maybe someone can direct me to the right place?

forki:
https://referencesource.microsoft.com/#mscorlib/system/string.cs,103 may
work as starting point

Dustin Moris Gorski <notifications@github.com> schrieb am Fr., 28. Sep.
2018, 15:40:

dustinmoris:
Ah nice, I didn't know about this type, because it is internal, so we cannot reuse it. Also this one has been tailored for smaller strings (because it's more frequently used from things like `string.Format` or `string.Concat`) and our's has higher default and max capacities because of the different nature of data that we are anticipating to deal with.

Anyways, looking at the [implementation](https://stackoverflow.com/questions/20029868/understanding-of-net-internal-stringbuildercache-class-configuration) it looks almost identical and the only useful addition we could perhaps maybe add would be the `GetStringAndRelease` method which is kind of nice.

Issue #311, opened 2018-10-01
[Proposal] Add encoding to content type response headers


dv00d00:
The source of this stuff is the techempower benchmark.

They require `text\html; charset=utf-8` content type header to be present.

It does affect browser behavior for nonascii html. I think we should include it by default for html and add a way to disable it somehow.

Other than that charset would be useful for all other response writers.

dv00d00:
before 
![image](https://user-images.githubusercontent.com/1162823/46318103-09a4b180-c5de-11e8-860a-4f9e2abe801a.png)

after
![image](https://user-images.githubusercontent.com/1162823/46318129-1f19db80-c5de-11e8-859e-31beba19775b.png)

dv00d00:
Actually I am not sure it should be optional

dv00d00:
Anyway, opened a pr #312 with the assumption that we wont need to omit charset in content type

dustinmoris:
This change is included in v.3.2.0, which is currently being cooked on the build server and then pushed to NuGet.

PR #312, opened 2018-10-01
Utf-8 charset

dv00d00:
As in #311


dv00d00:
This might be quite subtle change, not sure if it is backward compatible. Maybe it will still be worth adding a way to omit the charset, I will try to dig docs to see what is going on. 

So far a charset on html content feels mandatory, not so sure about plaintext and especially json

PR #317, opened 2018-10-20
Debug Microsoft.AspNetCore.TestHost 2.1

Tratcher:
https://github.com/aspnet/Hosting/issues/1537


Tratcher:
Interesting, StreamingTests passed when run by itself.

Tratcher:
So does PreconditionalTests

Tratcher:
Only reproduces with both StreamingTests and PreconditionalTests. XUnit defaults to running test classes in parallel. Do those tests share any resources?

I'll play with the Xunit settings to see if I get any more info.

dustinmoris:
I've commented all tests from `StreamingTests.fs` except the first one and all tests from `PreconditionalTests.fs` except the first one. This removes the noise and shows it requires only these two tests to deadlock each other.

Now in terms of sharing a common resource, this was obviously my first thought as well weeks ago and I don't see them sharing anything together and also it's worth remembering that if I'd change the `TestHost` package back to `2.0.*` again then it will work.

Here's the break down of what they do:

#### Test from `StreamingTests.fs`

```fsharp
[<Fact>]
let ``HTTP GET entire file with range processing disabled`` () =
    createRequest HttpMethod.Get Urls.rangeProcessingDisabled
    |> makeRequest
    |> isStatus HttpStatusCode.OK
    |> containsHeader false "Accept-Ranges"
    |> containsContentHeader false "Content-Range"
    |> hasContentLength 62L
    |> readText
    |> shouldEqual "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
```

#### Test from `PreconditionalTests.fs`

```fsharp
[<Fact>]
let ``HTTP GET with If-Match and no ETag`` () =
    createRequest HttpMethod.Get Urls.rangeProcessingDisabled
    |> addHeader "If-Match" "\"111\", \"222\", \"333\""
    |> makeRequest (None, None)
    |> isStatus HttpStatusCode.PreconditionFailed
    |> readBytes
    |> shouldBeEmpty
```

Tratcher:
Any way to make [this](https://github.com/giraffe-fsharp/Giraffe/blob/216a560947c70fc96280384c1bedc699857e785a/tests/Giraffe.Tests/Helpers.fs#L171) code async?

Tratcher:
t.Result is dangerous because it can block threads that are needed elsewhere. If enough threads get blocked then the system has trouble making progress. Xunit has a value [maxParallelThreads](https://xunit.github.io/docs/running-tests-in-parallel) is dynamic by default based on your CPU count. I wonder if in those CI environments the default is very low and it deadlocks easily.

dustinmoris:
> t.Result is dangerous because it can block threads that are needed elsewhere. If enough threads get blocked then the system has trouble making progress.

Yes I agree, but in this instance I think it would be still worth trying to understand why it's causing a deadlock, because a blocking operation like `t.Result` would normally only cause a sub optimal utilisation of threads, but it's not directly responsible for deadlocks. I would understand if the tests run perhaps a few ms longer, but a complete deadlock normally means that there is a shared resource which in this context causes two threads to lock each other by waiting for each other to release a resource. In my tests I don't have any shared state and with `TestHost` `2.0.*` there is also no such issue, so I'd guess that with `TestHost` `2.1.*` there has been some shared state introduced which now makes synchronous code like this (which might be inefficient but still perfectly fine) failing now.

> Xunit has a value maxParallelThreads is dynamic by default based on your CPU count. I wonder if in those CI environments the default is very low and it deadlocks easily.

Mhh.. that sort of matches what I thought when I mentioned the issue. If this setting can be changed then we should be able to set `maxParallelThreads` to `1` or `2` locally and reproduce this isse outside of CI systems too, which could help to further investigate the changes from TestHost 2.0 to 2.1, right?

> Any way to make this code async?

Done, let's see what the build will say now :)

Thanks for the help again!

dustinmoris:
Ok cool.. so making it async has fixed it. That's good, because that sort of is not massively surprising and gives us plenty of indication what type of issue we are dealing here with!

I guess I can change all tests to be async, which is a good idea anyway, but I would love to still find out why a blocking call like `t.Result` is causing a deadlock after the TestHost upgrade. I think this deserves further investigation, because it's probably not intended and we could learn something from it.

Tratcher:
The shared resource here is the threadpool itself. Here are the two TestServer changes in 2.1 that could be related:
[Response Stream Pipes](https://github.com/aspnet/Hosting/commit/b6dc00229d88d17023f0105c2e524e3cfab0888e#diff-cf9892ae7ae719b650226445cd46212e)
[HttpContext pattern](https://github.com/aspnet/Hosting/commit/82ccf4f06e749d2044f3adb88e870b431341274f#diff-cf9892ae7ae719b650226445cd46212e)

t.Result in this case is blocking two threads, one per test, and that appears to be enough to exhaust the number of threads allows by xunit.  All that need have changed is where makeRequest previously had completed synchronously with its callback invoked inline, to now where some part of that completion is being queued to the threadpool. It takes one free thread to process the threadpool queue and help complete the various async tasks.

Tratcher:
And yes, feel free to mess with maxParallelThreads to reproduce this elsewhere. I tried but couldn't get the settings file to apply.

PR #329, opened 2018-12-23
Add context and request http handlers

Zaid-Ajaj:
### Description 

This PR adds two convenient http handlers: `context` and `request` to the core module. These were really nice in `Suave` and I missed them in Giraffe because I didn't fully understand at first how to create a raw http handler that reads context information. They allow to create new http handlers based on information from the http context and http request. 

### Todo's:
- [x] Add `context` with docs
- [x] Add `request` with docs 
- [x] Add Release Notes
- [x] Add "Accessing HttpContext" to documentation
- [x] Follow the style guide
- [ ] Open an issue for discussion (we can discuss it in the PR)

### Documentation (section "Accessing HttpContext")

Reading information from the http context or the incoming http reqeust is sometimes needed to create new http handlers, that is why Giraffe has convenience http handlers like `context` and `request`. The former uses the HttpContext as input from which you can create a new http handler and the latter uses just the `HttpRequest` as input:

```fsharp
let handlerWithLogging = context (fun ctx ->
    let logger = ctx.GetService<ILogger>()
    logger.Information("From the context")
    text "Done working"
)

let echoRequestBody = 
    POST 
    >=> route "/echo"
    >=> request (fun req -> 
        use reader = StreamReader(req.Body)
        requestBody = reader.ReadToEnd()
        text requestBody
    )

### CODE REVIEW

**Comment by dustinmoris** [comment]
Yeah.. I have no strong opinions what the actual name should be, but I think it shouldn't be `context` or `request` because it would be too generic and very likely clash with code which someone would write elsewhere.

`fromContext` is slightly less long than `contextHandler`, but I'm not sure if it translates the concept behind the function as well as "handler":

```fsharp
let handlerWithLogging =
    contextHandler (
        fun ctx ->
            let logger = ctx.GetService<ILogger>()
            logger.Information("From the context")
            text "Done working"
    )

let handlerWithLogging =
    fromContext (
        fun ctx ->
            let logger = ctx.GetService<ILogger>()
            logger.Information("From the context")
            text "Done working"
    )
```

What do you think? Also agree, an async version would be nice to have too.

dustinmoris:
What about `handleContext` and `handleRequest`?

I think it reads fluently in code like your examples:

```fsharp
let handlerWithLogging =
    handleContext(
        fun ctx ->
            let logger = ctx.GetService<ILogger>()
            logger.Information("From the context")
            text "Done working")

let echoRequestBody =
    POST
    >=> route "/echo"
    >=> handleRequest(
        fun req ->
            use reader = StreamReader(req.Body)
            requestBody = reader.ReadToEnd()
            text requestBody)
```

Zaid-Ajaj:
Indeed `handleContext` and `handleRequest` do read very fluently, I like them! PR for the counter parts just pushed. 

> Even though these functions are quite simple, I am very pleased with them: now I don't have to remember how the signatures add up when creating custom handlers, thanks a lot for merging :smile:

PR #331, opened 2018-12-25
Add handleContextAsync and handleRequestAsync with docs

Zaid-Ajaj:
As discussed in #329,  this PR adds asynchronous counter parts for `handleContext` and `handleRequest` along with their docs and release notes. 

### Example use case
```fs
let echoContents (request: HttpRequest) = 
  task {
      use reader = new StreamReader(request.Body)
      let! contents = reader.ReadToEndAsync()
      return text contents 
  }

let webApp = POST >=> route "/echo" >=> handleRequestAsync echoContents
```

### DIFF
--- a/.psscripts/build-functions.ps1
+++ b/.psscripts/build-functions.ps1
@@ -309,8 +309,12 @@ function Install-NetCoreSdkFromArchive ($sdkArchivePath)
         Write-Host "Created folder '$env:DOTNET_INSTALL_DIR'."
         Expand-Archive -LiteralPath $sdkArchivePath -DestinationPath $env:DOTNET_INSTALL_DIR -Force
         Write-Host "Extracted '$sdkArchivePath' to folder '$env:DOTNET_INSTALL_DIR'."
-        $env:Path = "$env:DOTNET_INSTALL_DIR;$env:Path"
-        Write-Host "Added '$env:DOTNET_INSTALL_DIR' to the environment variables."
+        $updatedPath = "$env:DOTNET_INSTALL_DIR;$env:Path"
+        [Environment]::SetEnvironmentVariable("Path", $updatedPath, "Process")
+        [Environment]::SetEnvironmentVariable("Path", $updatedPath, "User")
+        [Environment]::SetEnvironmentVariable("Path", $updatedPath, "Machine")
+        Write-Host "Added '$dotnetInstallDir' to the Path environment variable:"
+        Write-Host $env:Path
     }
     else
     {
@@ -323,7 +327,8 @@ function Install-NetCoreSdkFromArchive ($sdkArchivePath)
         [Environment]::SetEnvironmentVariable("PATH", $updatedPath, "Process")
         [Environment]::SetEnvironmentVariable("PATH", $updatedPath, "User")
         [Environment]::SetEnvironmentVariable("PATH", $updatedPath, "Machine")
-        Write-Host "Added '$dotnetInstallDir' to the environment variables."
+        Write-Host "Added '$dotnetInstallDir' to the PATH environment variable:"
+        Write-Host $env:PATH
     }
 }
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ before_install:
   - curl https://packages.microsoft.com/config/ubuntu/14.04/prod.list | sudo tee /etc/apt/sources.list.d/microsoft.list
   - sudo apt-get update
   - sudo apt-get install -y powershell
-  - pwsh ./.psscripts/install-dotnet.ps1
+  - sudo pwsh ./.psscripts/install-dotnet.ps1
 
 script:
   - export PATH=/home/travis/.dotnetsdk:$PATH
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -247,44 +247,40 @@ let sayHelloWorld : HttpHandler =
         }
 ```
 
-#### Convenience functions
+#### handleContext
 
-With version 3.5.0 and onwards Giraffe exposes additional convenience functions which can be used to create new `HttpHandler` functions which require access to the `HttpContext` or `HttpRequest` object without having to define the full verbose implementation as shown above.
+Starting with version 3.5.0 and onwards Giraffe exposes an additional convenience function which can be used to generate new handler functions which only require access to the `HttpContext` object without having to define the full verbose implementation as shown above.
 
-The `handleContext` and `handleRequest` functions, along with their asynchronous counter parts `handleContextAsync` and `handleRequestAsync` can be used like this:
+The `handleContext` function can be used like this:
 
 ```fsharp
-let handlerWithLogging =
+let handlerWithLogging : HttpHandler =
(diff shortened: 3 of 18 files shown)


dustinmoris:
Hi,

Thanks for submitting this PR and also fixing the docs as part of it :).

I've been thinking about these new functions this morning and would like to propose a slight change. I am extremely particular and want to ensure that the Giraffe library makes overall sense in how things have been designed and to keep things consistent as new functionality gets added.

With that in mind I am not super happy about having two different functions which do essentially the same thing: `handleContext` vs `handleContextAsync`.

What bothers me is that in Giraffe every other handler doesn't have an `Async` equivalent, because all `HttpHandler` functions are async by definition. This brings me to the second point which is that `handleContextAsync` now suggests that `handleContext` is not asynchronous which is not true. If I call `text "Hello World"` from `handleContext` then the entire operation will also be executed asynchronously by ASP.NET Core, because the end result is still a `Task<HttpContext option>`.

I was thinking it would be better if we could have a single `handleContext` function which can be defined with or without the `task {}` CE. It would be more consistent with the programming model in Giraffe and also set the correct expectations.

The modified solution would look like this:

```fsharp
let handleContext (contextMap : HttpContext -> HttpFuncResult) : HttpHandler =
        task {
            match! contextMap ctx with
            | Some c ->
                match c.Response.HasStarted with
                | true  -> return  Some c
                | false -> return! next c
            | None      -> return  None
        }
```

When used:

```fsharp
let handlerWithLogging : HttpHandler =
    handleContext(
        fun ctx ->
            let logger = ctx.GetService<ILogger>()
            logger.LogInformation("From the context")
            ctx.WriteTextAsync "")
```

Or by performing additional async operations via the `task {}` CE:

```fsharp
let handlerWithLogging : HttpHandler =
    handleContext(
        fun ctx ->
            task {
                let logger = ctx.GetService<ILogger>()
                logger.LogInformation("From the context")
                // Do more async stuff
                return! ctx.WriteTextAsync "Done working"
            })
```

I think this feels like a perhaps slightly cleaner implementation, however it has two drawbacks:

- You cannot return a `HttpHandler` anymore. It has to be a `HttpFuncResult` (`Task<HttpContext option>`), which means that instead of writing `text "Hello World"` one has to write `ctx.WriteTextAsync "Hello World"`. This is not a huge problem IMHO, because every `HttpHandler` function has an equivalent `ctx` extension method (this is what they are all calling under the covers).
- The second issue is that we can't have a `handleRequest` convenience method anymore. Here we wouldn't be able to provide one function for both scenarios - non Task CE and Task CE versions. Again I think this is not a big issue because the `handleContext` function can just be used instead, since the `HttpRequest` object is literally a direct property of `ctx`: `ctx.Request`.

I don't think both issues are a big deal, but I thought it's worth mentioning and would love to hear your thoughts on this!

dustinmoris:
I've pushed the proposed change to this branch so you can pull the latest and have a look at my suggested implementation. We can still roll back my change or make more modifications, this is purely to give you more context.

Zaid-Ajaj:
Hi @dustinmoris, thanks a lot for the detailed response, here is my reasoning for this 

> I am extremely particular and want to ensure that the Giraffe library makes overall sense in how things have been designed and to keep things consistent as new functionality gets added.

As a library author myself, I understand this all too well. Avoiding feature bloat and keeping a library consistent can be really tricky. This will confuse some users and they will open issues like "Where is the async version of `<handler>`?" while they, in fact, are all asynchronous. 

> With that in mind I am not super happy about having two different functions which do essentially the same thing: handleContext vs handleContextAsync. 

> I was thinking it would be better if we could have a single handleContext function which can be defined with or without the task {} CE. It would be more consistent with the programming model in Giraffe and also set the correct expectations.

I agree with you to some extent, especially from the consistency point of view of the library but there is a little problem if you look from the user perspective: the Giraffe model provides really *simple* ways of transforming the response, i.e. `GET >=> route "/" >=> text "Index"` but if things become a little bit more complex (the need to use and `Task<'t>`) then the user has to start using *low-level* contructs and maniuplate the `HttpContext` directly with methods provided by ASP.NET Core. 

To make a long story short: I think it is quite the leap from using the existing `HttpHandlers` (simple) to implementing your own handlers with context and next functions (low-level) just because you now need to use `task`. I was thinking of how to explain the concepts of the Giraffe model to my colleagues and thought of the transition in difficulty, which is why I wanted to add the task returning versions of `handleContextAsync` and `handleRequestAsync` as being in the middle of simple handlers and low-level middleware handlers: 

> Keep in my when I say "low-level" I mean for my colleagues: beginners both in F#, web dev and the Giraffe model of web dev. 

### Step 1: combine simple handlers
```fs
GET >=> route "/" >=> text "Index"
GET >=> route "/about" >=> htmlView aboutPage
```
### Step 2: parameterize request by path:
```fs
GET >=> routef "/album/%d" (fun id -> text (sprintf "You requested album id = %d" id))
```
### Step 2: access to all request information:
```fs
POST 
  >=> route "/echo" 
  >=> handleRequest (readRequestContent >> text) 
```
### Step 4: complex behaviour with context
```fs
let getAlbumById (albumId: int) (db: IDatabase) =
  task {
      let! albumResult = db.getAlbumById albumId
      match albumResult with
      | DbError -> return setStatusCode 500 >=> text "Internal database error"
      | AlbumNotFound -> return setStatusCode 404 >=> text "Album was not found"
      | Album album -> return statusCode 200 >=> setJsonContentType >=> text (Json.serialize album)
  }

GET
  >=> routef "/albums/%d" (
    fun albumId -> 
     handleContextAsync(
       fun context ->
          let db = context.GetService<IDatabase>()
          getAlbumById albumId db))
```
IMHO, all these examples are of the same level of difficulty (read: same level of abstraction) because they all rely on existing handlers, just composing them differently and with the last example you are able to do a lot before delving into the inner workings of how these handlers are implemented, you also re-use the handlers that allow to transform the response without resorting to functions from the `HttpContext`. 

# Conculsion

The changes you proposed are good and I think they should stay like that, how "low-level" something is, is very subjective and it is just how I thought of it. Giraffe has great documentation that people can always refer to it if they need some more complex behaviour, if these helpers aren't added in the core library, it would be really nice to add them to the docs as convenience helpers that don't have a place in the big picture of the library

Zaid-Ajaj:
Forget to close, since there is nothing to do from here on. As concluded, your proposed changes should stay as is, thank you!

dustinmoris:
@Zaid-Ajaj Sorry for the delayed response. I wanted to get back and say thanks for the help and also try to address

> if these helpers aren't added in the core library, it would be really nice to add them to the docs as convenience helpers that don't have a place in the big picture of the library

Adding this to the docs is definitely something I can easily do, but I was thinking if there was something more I could to to help you and I thought perhaps something like `Giraffe.Experimental` like Suave has would be a module where I could add these handlers. What do you think?

Zaid-Ajaj:
Yes I really like the `Giraffe.Experiemental` idea as playground for such functionality, though I proposed a different functionality in #333 with the same purpose in mind, looking forward to hearing your feedback

Issue #333, opened 2018-12-28
Feature request: execute and resolveService


Zaid-Ajaj:
As we had discussed some ideas in #331, my main point was to hide implementation details of `HttpHandler` from the end user unless he/she really wants to build middleware themselves. My propostion of adding `handleContextAsync` didn't really make sense in the big picture of the library so here is an alternative that just might:

Let's assume the user had the function:
```fs
getAlbums : IMusicStore -> Task<Album list>
```
I want it to be really really simple to expose this through HTTP without thinking too much, I need:
- A way to get the *input* (`IMusicStore` in this case coming from IoC container) 
- A way to map the *output* (`Album list` as HTTP response)

Here is where `execute` and `resolveService` come into play with these types:
```fs
execute : Task<T> -> (T -> HttpHandler) -> HttpHandler
resolveService<'T>: ('T -> HttpHandler) -> HttpHandler
```
you can already tell where I am going with this :smile: 

First thing I have to think about is how to map the output of the function through Http, maybe map to JSON or return a Razor view etc.
```
mapAlbumsToJson : Album list -> HttpHandler = (* serialize to JSON for example *) 
``` 
then all I need to do it extract `IMusicStore` from the IoC container and chain the functions:
```fs
GET 
  >=> route "/albums"
  >=> resolveService<IMusicStore> (
      fun musicStore -> execute (getAlbums musicStore) mapAlbumsToJson
   )
```
if more parameters were needed like `albumId: int` then we would have used `routef` and passed in the path parameter to the function in question so no need for extra HTTP handler for extracting input. 

The names are of course subject to change (`execute` is probably too generic) but the I hope you get my idea. @dustinmoris I also like the `Giraffe.Experiemental` proposition but for these I would like to add them to `Core.fs` 

Let me know what you think!

dustinmoris:
Hi, I think both handlers sound reasonable. Feel free to send a new PR for those!

PR #334, opened 2018-12-29
Add resolveService<'t> and mapTask<'t>

Zaid-Ajaj:
Addresses #333, adds: `resolveService<'t>` and `mapTask<'t>` (instead of `execute`) with docs, release notes etc.

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -276,6 +276,58 @@ let handlerWithLogging2 : HttpHandler =
             })
 ```
 
+### resolveService<'t> and mapTask<'t>
+
+> Applies to Giraffe v3.6.0+
+
+Allows to resolve a service of type `'t` registered in the IoC container of ASP.NET Core's dependency injection machanism and maps the resolved service into another `HttpHandler`
+
+Example 1: map the result of a (sync) function to JSON using a dependency:
+```fsharp
+type IAlbumStore = 
+    abstract getAllAlbums : () -> Album list
+
+GET 
+  >=> route "/albums"
+  >=> resolveService<IAlbumStore>(fun albumStore -> json (albumStore.getAllAlbums()))
+```
+Example 2: combine and resolve multiple services: 
+```fsharp
+type IAlbumStore = 
+    abstract getAllAlbums : () -> Album list
+    abstract albumsForUser : User -> Album list
+
+type IUserStore = 
+    abstract getCurrentUser : () -> User 
+
+GET 
+  >=> route "/my-albums"
+  >=> resolveService<IUserStore>(fun userStore ->
+        resolveService<IAlbumStore>(fun albumStore ->
+            let currentUser = userStore.getCurrentUser()
+            let albums = albumStore.albumsForUser currentUser
+            json albums
+        )
+    )
+```
+Example 3: resolve service and maps their async (task-based) functions into a `HttpHandler`
+```fsharp
+type IAlbumStore = 
+    abstract getAllAlbums : () -> Task<Album list>
+    
+GET 
+  >=> route "/albums"
+  >=> resolveService<IAlbumStore>(
+      fun albumStore -> 
+        let albumsTask = albumStore.getAllAlbums() 
+        let albumListToJson (albums: Album list) = json albums 
+        mapTask albumsTask albumListToJson)
+
+// Simplified 
+GET 
+  >=> route "/albums"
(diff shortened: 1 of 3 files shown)


NinoFloris:
@Zaid-Ajaj please don't use task{} here, no need to pay for going async in these cases

Zaid-Ajaj:
@NinoFloris Pay for what exactly, an extra function call? How significant is this?

dustinmoris:
Hey, so I've let this PR sit for a while, but it's time to get this wrapped up.

I like how you made resolving a dependency very straight forward with the helper `resolveService`, but the `mapTask` function I can't get my head around properly. I know what it is doing, but I'm not sure if it only exists for the use in combination with `resolveService` or if it has a usage beyond that.

Apologies that some of my feedback comes a bit delayed, but I like to pull a PR locally and play with new APIs for a while to get a feeling for different use cases, how people might want to apply it and what someone could do with it beyond the original intent.

With that in mind I've made a small change in my local version of it, where I swapped the two input parameters in order to allow `|>` piping of a service into a `mapTask` function:

```fsharp
let mapTask<'T> (mapper: 'T -> HttpHandler) (inputTask: Task<'T>) : HttpHandler =
        task {
            let! extracted = inputTask
            let resultHandler = mapper extracted
            return! resultHandler next ctx
        }
```

Now I could do something like this:

```fsharp
let someHandler : HttpHandler =
    GET >=>
    route "/albums" >=>
    resolveService<IAlbumsService> (
        fun albumsService ->
            albumsService.GetAlbumsAsync()
            |> mapTask (fun albums -> text "Hello World")
    )
```

Now that works quite neat, but I'm trying to picture when would I use that instead of just defining a normal `HttpHandler` function? The original intent is to hide some implementation details and make the API look simpler, but as soon as `resolveService<'T>` returns a service which has async methods then I think it starts becoming messy.

I'm not sure which one a beginner would find more straightforward between these two solutions:

```fsharp
let someHandler : HttpHandler =
    GET >=>
    route "/albums" >=>
    resolveService<IAlbumsService>(
        fun albumsService ->
            resolveService<IUserService>(
                fun userService ->
                    userService.GetUserAsync()
                    |> mapTask (fun user -> 
                        albumsService.GetAlbumsByUserAsync user
                        |> mapTask (fun albums -> text "Hello World")
                    )
            )
    )
```

vs.

```fsharp
let someHandler : HttpHandler =
    GET >=>
    route "/albums" >=>
    handleContext (fun ctx ->
        task {
            let userService = ctx.GetService<IUserService>()
            let albumsService = ctx.GetService<IAlbumsService>()
            let! user = userService.GetUserAsync()
            let! albums = albumsService.GetAlbumsByUserAsync user
            return! ctx.WriteTextAsync "Hello World"
        }
    )
```

This makes me believe that as soon as someone has to resolve more than one dependency and potentially even a dependency which will execute some async methods then these handlers will result in a lot of function nesting which means that defining a normal handler could be the easier way at this point?

Zaid-Ajaj:
> I know what it is doing, but I'm not sure if it only exists for the use in combination with `resolveService` or if it has a usage beyond that.

Indeed, I didn't have other use cases for it in mind, it just makes simple when working with `resolveService` but is is worth noting that `mapTask` is not intended for use with nesting but instead it should be used at the very end:

The idea is to seperate *using* the dependencies from *resolving* them to make functions that require these dependencies easily unit-testable outside of the Giraffe HTTP pipeline
```fs
// pure function, no http-specific stuff and easily unit-testable outside of Giraffes' Http pipeline
let topTenAlbums (users: IUserService) (albumStore: IAlbumStore) = 
    task {
       let! currentUser = users.getCurrent()
       let! userAlbums =  albumStore.getForUser currentUser
       return userAlbums |> List.take 10 
    }

let someHandler : HttpHandler =
    GET >=>
    route "/albums" >=>
    resolveService<IAlbumStore>(
        fun albums ->
            resolveService<IUserService>(
                fun users-> mapTask (topTenAlbums users albums) json
            )
    )
```
even though you will have the nesting of `resolveService` which I didn't like that much. The way to solve it is by overloading the `resolveService` function to take multiple type parameters for multiple services which can only be done with static functions from a static class. I thought this wouldn't make sense in the core of Giraffe so I built a library in the meanwhile, [Giraffe.GoodRead](https://github.com/Zaid-Ajaj/Giraffe.GoodRead) for this exact purpose along with nice helpers to work with task/async. 

I am happy to accept changes to and merge the PR, but if you think the API doesn't make sense I will be happy to close it as well since these concerns can easily solved with a library, what do you think :smile:?

PR #337, opened 2019-01-05
Doc - remove tangential task library info, fix a few typos

chadunit:
It seemed out of place to have an explanation of how to use an external library in a non-Giraffe context, and a single link to the library should be sufficient for reference, so removed the extra details. Also fixed a few typos throughout the document.


DOCUMENTATION.md
```diff
@@ -172,25 +172,6 @@ let personHandler =
...
-        let! contents = reader.ReadToEndAsync()
-
-        // do something with contents
-
-        return contents
-    }
-```
-
-For more information please visit the official [TaskBuilder.fs](https://github.com/rspeele/TaskBuilder.fs) GitHub repository.
-
```
slang25: How about leaving this bit in, just in case the reader does want to know more:

```
The `task {}` CE is an independent project maintained by [Robert Peele](https://github.com/rspeele), for more information please visit the official [TaskBuilder.fs](https://github.com/rspeele/TaskBuilder.fs) GitHub repository.
```

I know that opening part is duplicating what is above a bit (i.e. task{} is not from Giraffe), but it's worth reiterating, and it gives credit to the creator.

PR #339, opened 2019-01-28
Fix nested subRouteCi case sensitivity

### LINKED ISSUE #338: subRouteCi not case insensitive
I've found that the case insensitive variant of `subRoute` still requires the route as a whole to match the case exactly as is defined in the code in case it has nested path filters.

A router set up like this would only get matched all the way to the `text` if the request uses all lowercase path:
```fsharp
subRouteCi "/foo" (route "/bar" >=> text "hi!")
```
In cases like `"/FOO/bar"`, the `subRouteCi` itself works, but `route /"bar"` won't get matched.

I have a runnable repro as a unit test here:
https://github.com/jbrestan/Giraffe/blob/subrouteci-repro/tests/Giraffe.Tests/RoutingTests.fs#L1084

It looks like the case information is lost between correctly filtering the request path and storing it in the `HttpContext`.
https://github.com/giraffe-fsharp/Giraffe/blob/bdc02592cb23f6d63c188e603d2050b9e688f270/src/Giraffe/Routing.fs#L321-L323
The context then contains the path fragment from the code, not from the request, and the nested filters do case sensitive comparison on that, which will fail unless it's an exact match.
https://github.com/giraffe-fsharp/Giraffe/blob/bdc02592cb23f6d63c188e603d2050b9e688f270/src/Giraffe/Routing.fs#L29

Saving the path fragment from the request instead of the string from code should make the nested filters work without change to the fast exact-match validation on the value in context.

I can prepare a PR for this fix if you're OK with the proposed solution.

jbrestan:
Fixes #338 

Not sure about the naming convention for unit tests

### DIFF
--- a/src/Giraffe/Routing.fs
+++ b/src/Giraffe/Routing.fs
@@ -319,8 +319,12 @@ let subRoute (path : string) (handler : HttpHandler) : HttpHandler =
 /// A Giraffe `HttpHandler` function which can be composed into a bigger web application.
 ///
 let subRouteCi (path : string) (handler : HttpHandler) : HttpHandler =
-    routeStartsWithCi path >=>
-    SubRouting.routeWithPartialPath path handler
+    fun (next : HttpFunc) (ctx: HttpContext) ->
+        let nextPartOfPath = SubRouting.getNextPartOfPath ctx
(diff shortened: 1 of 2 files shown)


dustinmoris:
Thanks for the fix and also adding tests for it!!

jbrestan:
Np, happy to help, thanks for an awesome library! :)

PR #340, opened 2019-02-18
Update README.md

vivainio:
Remove weird stuff from "dotnet add package" commands


dustinmoris:
Thanks, I think I copied these commands from the official MS website, but I guess this is not really needed and makes it look more confusing to starters.

PR #344, opened 2019-03-07
Fix a typo in the accessibility handler for `labelledby`

### LINKED ISSUE #343: Accessibility property is misspelled
The value for the `aria-labelledby` property is misspelled and is producing invalid markup. It only has one `l` and should have two.

https://github.com/giraffe-fsharp/Giraffe/blob/c9e135d7c26d6418ac3eb7e1f2ab5b949ad08a7d/src/Giraffe/GiraffeViewEngine.fs#L492

baronfel:
Closes #343 by fixing the typo.


dustinmoris:
Thank you!

Issue #341, opened 2019-02-20
routeStartsWithf missing?


NinoFloris:
I'm working on Saturn routing right now and hitting a snag around the fact the tryMatchInput regex ends in $, could we add a flag to opt out of that? Then we could add some route functions (routeStartsWithCif and routeStartsWithf) on top to expose that

dustinmoris:
Yes I suppose we could add a `(exactMatch : bool)` flag to the `tryMatchInput` function. Feel free to submit a PR!

dustinmoris:
Out in 4.0.0

PR #353, opened 2019-04-26
routeStartsWithf + routeStartsWithCif

dv00d00:
Added support for regular expression match mode
Added two extra routing extensions which are using it
Added some basic tests

Addresses #341

This is draft, comments are welcome

Edit: after looking at it for a while, I am pleased with the current state of pr, or I lost the ability for self-criticism

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -30,6 +30,18 @@ For more information please see [issue #347](https://github.com/giraffe-fsharp/G
 - Fixed `routef` to not match more than one URL path segment.
 - Fixed the `_ariaLabelledBy` attribute in the `GiraffeViewEngine`
 - Fixed case insensitive route handlers on Ubuntu
+- Renamed `tryMatchInput` to `tryMatchInputExact`
+- Added new version of `tryMatchInput` which accepts `MatchSettings` record:
+
+    ```fsharp
+    type MatchMode =
+        | Exact                // Will try to match entire string from start to end.
+        | StartsWith           // Will try to match a substring. Subject string should start with test case.
+        | EndsWith             // Will try to match a substring. Subject string should end with test case.
+        | Contains             // Will try to match a substring. Subject string should contain test case.
+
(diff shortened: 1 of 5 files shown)


dustinmoris:
This looks great! The new changes are very subtle but effective. I like it a lot.

dv00d00:
Naming is hard though. I kept old tryMatchInput as it was public and added a replacemnt with options, wthout thinking much of its name 😄 . And such tryMatchInputOptions was born. I don't know how to fix it though, I can think of an overaload as a probable solution

dv00d00:
Ok, overloading works on toupled methods on a class, not sure how to get away from breaking binary compatibility here and do not introduce a new name

dv00d00:
I've done minor renaming + local perf tuning which would not affect much, except only it shall not make things worse than they were before refactoring

dustinmoris:
Thanks for the extra changes. I was thinking about the naming too and I think `tryMatchInput` should be renamed to `tryMatchInputExact` as that is what it;s really doing with the only option being case sensitivity, whereas `tryMatchInput0` is what gives full flexibility and should therefore probably just be named `tryMatchInput`. This would be a breaking change, but the next upcoming release already has other minor breaking changes like this. If you agree then I will document it accordingly in the release notes. What do you think?

dv00d00:
Sounds great @dustinmoris 
I will commit the renames and try to update docs where apropriate

dv00d00:
Once we have started break things, maybe reorder arguments list and put string to the last position?
e.g.
Before
```fsharp
let tryMatchInput (format : PrintfFormat<_,_,_,_, 'T>) (input : string) (options : MatchOptions) 
```
After
```fsharp
let tryMatchInput (format : PrintfFormat<_,_,_,_, 'T>) (options : MatchOptions) (input : string) 
```

To allow easier partial application?

dustinmoris:
Thanks for this PR! I've also swapped the arguments as I think that's a good suggestion!

Issue #351, opened 2019-04-13
With Giraffe 3.6.0, BindJsonAsync in task CE yields runtime error message "Synchronous operations are disallowed. Call ReadAsync or set AllowSynchronousIO to true instead."


cpbotha:
In a `POST` handler task CE, the following code yields the error message in the title of this issue:

```fsharp
let! body = ctx.BindJsonAsync<NewCommentBody>()
```

However, by instead following the lead of `BindXmlAsync`'s implementation, the following code works perfectly:

```fsharp
let serializer = ctx.GetJsonSerializer()
let! bodyText = ctx.ReadBodyFromRequestAsync()
let body = serializer.Deserialize<NewCommentBody> bodyText;
```

You can compare the implementations of  `BindJsonAsync` and `BindXmlAsync` here: https://github.com/giraffe-fsharp/Giraffe/blob/master/src/Giraffe/ModelBinding.fs#L241

slang25:
I've managed to recreate this, it looks like we would need Json.NET to add a `DeserializeAsync` entry point (they have async readers already) https://github.com/JamesNK/Newtonsoft.Json/issues/1193

Microsoft are also bringing a new fast and async library for working with JSON which would be a good idea to offer too. I'm happy to work on this when it matures.

dv00d00:
Should not this be working fine with Utf8JsonSerializer ?

dustinmoris:
Out in 4.0.0

PR #354, opened 2019-04-26
Removes blocking calls to http stream when using Newtonsoft json

dv00d00:
Untill true async apis are available from Newtonsoft itself, this is probably the way to go to be able to work with .net core 2.2+

Sumary of changes: do not allow newtonsoft json classes touch raw http stream, cache in memory stream first.

Addresses #351


dv00d00:
I am not sure about the call to get stream length. This might or might not throw. I am keen to disable it unless somebody knows for sure if it is safe to use it.

src/Giraffe/Serialization.fs
```diff
@@ -92,11 +94,11 @@ module Json =
                 |> Encoding.UTF8.GetBytes
 
             member __.SerializeToStreamAsync (x : 'T) (stream : Stream) =
-                use sw = new StreamWriter(stream, Utf8EncodingWithoutBom, DefaultBufferSize, true)
-                use jw = new JsonTextWriter(sw)
-                let sr = JsonSerializer.Create settings
-                sr.Serialize(jw, x)
-                Task.CompletedTask
+                let memoryStream = new MemoryStream()
+                let streamWriter = new StreamWriter(memoryStream, Utf8EncodingWithoutBom)
```
dustinmoris: Should we be using the `use` keyword for the `StreamWriter` as it is disposable (same for the `JsonTextWriter` and the readers down below)?

dv00d00: I was sure this is for disposal of underlying stream, in this case the stream is a resize array, but I will check

dv00d00: Good catch, it flushes buffer on dispose, will introduce `use`s where they were

slang25: Do those flushes need to be async too?  See here: https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#always-call-flushasync-on-streamwriters-or-streams-before-calling-dispose

dv00d00: Yeah, needs to be adressed, great catch!

dv00d00: There is no need in calling flush async here because writes to memory stream are synchronous anyway. 
I've added manual flush before copy to the output though

slang25: Cool, I thought that might be the case, I sometimes get in a middle about what should be flushed when, especially when there are multiple streams involved

TheAngryByrd:
Sorry to necro this, but the trade off here now is using the `jsonChunked` function is allocating in memory, meaning larger responses have a giant buffer in memory rather than writing directly to the stream.  I would think this is the opposite of what the `jsonChunked` is trying to accomplish.

Issue #355, opened 2019-04-28
Add an example of a button onclick attribute that calls a javaScript function


ScottHutchinson:
The section at https://github.com/giraffe-fsharp/Giraffe/blob/master/DOCUMENTATION.md#common-view-engine-features should have an example of a button onclick attribute that calls a javaScript function.

PR #356, opened 2019-04-28
first pass of documentation around consuming JS in event handlers in Giraffe

baronfel:
Addresses #355 by adding documentation around wiring up Javascript event handlers via various methods.  Is this desired at all? Should I include 'learn more at...' links to Mozilla Developer Network pages around script-inclusion?

Related: do we want to html-encode these attributes? or should they be treated as `rawText` is, unencoded?

Should one of the examples show the passing-in of the event payloads and/or usage of `this` being auto-set to the element that triggered the event? See https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Event_handlers#Event_handler's_parameters_this_binding_and_the_return_value for more details of that mechanism.


dustinmoris:
> Should I include 'learn more at...' links to Mozilla Developer Network pages around script-inclusion?

Not necessary IMHO, but adding a single link to other helpful resources doesn't hurt either :)

> Related: do we want to html-encode these attributes? or should they be treated as rawText is, unencoded?

Hmm.. this is a very good question. My initial instict says that the JavaScript event handlers perhaps shouldn't get HTML encoded...

I'm trying to think of a scenario where someone would want to put user generated input into such an event handler and I cannot think of any. This is more common for HTML elements which have been dynamically generated from JavaScript itself - in which case it's already happening at JS level and not at the initial page render which is done by Giraffe. The other use case I could potentially think of is if someone was to build a CMS like system where user generated input would get rendered into a JavaScript event handler in Giraffe, but in this case the author of such system could/would have to assess the risk themselves and potentially encode it as an extra step.

So overall I think we should probably be ok by not html encoding these attributes, but I'm open to other opinions!

> Should one of the examples show the passing-in of the event payloads and/or usage of this being auto-set to the element that triggered the event?

I would have thought that most people will probably know how this works, but what do you think @ScottHutchinson ?

ScottHutchinson:
@dustinmoris Re: Should one of the examples show the passing-in of the event payloads and/or usage of this being auto-set to the element that triggered the event?

I think that would be a helpful addition. So often, .NET documentation includes only the simplest examples (e.g., without any arguments), even though some very common use cases require something a bit more complex. Since Giraffe is a front-end alternative to ASP.NET Core MVC, about which there are several long books available, anything the documentation can do to fill in the gaps would make it more likely that someone could succeed in using it for a real-world application.

dustinmoris:
> I think that would be a helpful addition. So often, .NET documentation includes only the simplest examples (e.g., without any arguments), even though some very common use cases require something a bit more complex.

Ok cool, I'll merge what we've got here for now as I'm prepping a new release, but I'll keep the original issue open so we can add this additions as well in order to be complete!

PR #364, opened 2019-07-12
Lower Json.NET version requirement to 11.0.2

### LINKED ISSUE #350: Lower Newtonsoft requirements ?
Hello, I am taking a look to upgrade `SAFE-Bookstore` repo to use `Fable 2.2` and `Fable.Core v3`. 

While doing that I discovered that [Microsoft.NET.Sdk.Functions](https://www.nuget.org/packages/Microsoft.NET.Sdk.Functions/) pinned `Newtonsoft.Json` to `11.0.2`. So I lower the version required by `Thoth.Json.Net` to be `>=  11.0.2` so people using `Microsoft.NET.Sdk.Functions` can use the new major version of my library.

However, it seems like [Giraffe.fsproj](https://github.com/giraffe-fsharp/Giraffe/blob/master/src/Giraffe/Giraffe.fsproj#L51) is asking for version superior to `12.0.*`. So people will not be able to use the latest version of Giraffe if they also use Azure.

Please note, that I know almost nothing about Giraffe and Azure, but because I had the same situation with `Thoth.Json` I wanted to mention the situation so you can take a look at it and decide if something needs to change or no.

cc @nojaf because we worked together on the situation for Thoth.Json and perhaps you use Giraffe too.
cc @forki because perhaps have an opinion about it as one of the SAFE Bookstore repo

slang25:
Fixes #350

I am struggling to use Giraffe with Azure Functions, and there appears to be no downside to this. It appears that aspnetcore mvc indirectly targets 11.0.2, like azure functions and a lot of Microsoft packages.

We could target 9.x, however it feels appropriate to target 11.x as that is the earliest version to support netstandard2.0, and 11.0.2 has an F# relevant fix: "Fix - Fixed serializing F# enums".

I'm hoping we can get this into a 4.0.0 release?

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -30,6 +30,7 @@ For more information please see [issue #347](https://github.com/giraffe-fsharp/G
 - Fixed `routef` to not match more than one URL path segment.
 - Fixed the `_ariaLabelledBy` attribute in the `GiraffeViewEngine`
 - Fixed case insensitive route handlers on Ubuntu
+- Changed minimum version of `Newtonsoft.Json` to `11.0.2`. This allows Giraffe to be compatable with Azure Functions.  
 
 ## 3.6.0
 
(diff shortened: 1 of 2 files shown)


slang25:
Rebased against develop.

dustinmoris:
This also looks good to me. Again, do you know of any reason why not merge at this stage? Thanks for the work!

PR #365, opened 2019-07-14
Assortment of minor infrastructure changes

slang25:
Here is a collection of minor changes, including:
- Using a more recent .NET Core SDK
- Updating packages and using specific version numbers for non-public dependencies
- Use Microsoft package for reference assemblies and removing old workarounds
- Getting tests running under mono


dustinmoris:
This looks good to me. Any reason why we shouldn't merge this that I should be aware of?

PR #369, opened 2019-08-07
Fixing jsonChunked

### LINKED ISSUE #368: jsonChunked does not follow RFC
The `jsonChunked` function does not follow the RFC for `Transfer-Encoding: chunked`. 

jsonChunked calls `SerializeToStreamAsync` on the current JsonSerializer.

https://github.com/giraffe-fsharp/Giraffe/blob/6e33155d3f0929af52886a7302fbd12a07ff9cb7/src/Giraffe/ResponseWriters.fs#L156-L163

Which `SerializeToStreamAsync`'s implementation is writing directly to the output stream

https://github.com/giraffe-fsharp/Giraffe/blob/7981279dfef1432b271975449c8fcb5935cf977c/src/Giraffe/Serialization.fs#L95-L99

---

If you use `curl` to hit an endpoint using `jsonChunked` you'll get an error

```
* Illegal or missing hexadecimal sequence in chunked-encoding
* stopped the pause stream!
* Closing connection 0
curl: (56) Illegal or missing hexadecimal sequence in chunked-encoding
```

Similarly browsers and javascript libraries don't know how handle this response. The problem is it doesn't follow the RFC for `Transfer-Encoding: chunked`.

The [RFC BNF](https://tools.ietf.org/html/rfc2616#page-25) is shown below

```
       Chunked-Body   = *chunk
                        last-chunk
                        trailer
                        CRLF

       chunk          = chunk-size [ chunk-extension ] CRLF
                        chunk-data CRLF
       chunk-size     = 1*HEX
       last-chunk     = 1*("0") [ chunk-extension ] CRLF

       chunk-extension= *( ";" chunk-ext-name [ "=" chunk-ext-val ] )
       chunk-ext-name = token
       chunk-ext-val  = token | quoted-string
       chunk-data     = chunk-size(OCTET)
       trailer        = *(entity-header CRLF)
```

An example of this is [from MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding#Chunked_encoding)

```
HTTP/1.1 200 OK 
Content-Type: text/plain 
Transfer-Encoding: chunked

7\r\n
Mozilla\r\n 
9\r\n
Developer\r\n
7\r\n
Network\r\n
0\r\n 
\r\n
```

If we're going to keep the interface of writing to a Stream for the json serializer, we're going to have to implement some type of Stream that handles writing this format to the HttpContext's response stream.

TheAngryByrd:
Closes #368 

- ~~This follows the RFC (partly) for using chunked transfer encoding.~~
- Ok this now defers to asp.net core's chunking.  [Setting  `Transfer-Encoding` to `chunked` manually](https://github.com/aspnet/AspNetCore/blame/728110ec9ee1b98b2d9c9ff247ba2955d6c05846/src/Servers/Kestrel/test/InMemory.FunctionalTests/ChunkedResponseTests.cs#L494) tells asp.net to defer to the caller to handle chunking manually.   Removing this fixes the issue.  
- This also replaces `MemoryStreams` with [`RecyclableMemoryStream`](https://www.philosophicalgeek.com/2015/02/06/announcing-microsoft-io-recycablememorystream/) to help with alleviating Large Object Heap pressure for large json serialization.


src/Giraffe/ResponseWriters.fs
```diff
@@ -157,9 +160,34 @@ type HttpContext with
...
+            }
             if this.Request.Method <> HttpMethods.Head then
-                let serializer = this.GetJsonSerializer()
-                do! serializer.SerializeToStreamAsync dataObj this.Response.Body
+                let pool = ArrayPool<char>.Shared
+                let buffer = pool.Rent 4096
+                try
+                    use memoryStream = recyclableMemoryStreamManager.GetStream()
+                    let serializer = this.GetJsonSerializer()
+                    do! serializer.SerializeToStreamAsync dataObj memoryStream
```
TheAngryByrd: Unfortunately this design does suffer from copying to two separate `RecyclableMemoryStreams`

src/Giraffe/ResponseWriters.fs
```diff
@@ -9,10 +9,13 @@ open Microsoft.AspNetCore.Http
 open Microsoft.Net.Http.Headers
 open FSharp.Control.Tasks.V2.ContextInsensitive
 open Giraffe.GiraffeViewEngine
+open Microsoft.IO
+
 
 // ---------------------------
 // Helper functions
 // ---------------------------
+let recyclableMemoryStreamManager = RecyclableMemoryStreamManager()
```
TheAngryByrd: I possibly could move this to either a Singleton with IoC or some other location since we are using it twice now.

src/Giraffe/ResponseWriters.fs
```diff
@@ -157,9 +160,34 @@ type HttpContext with
...
+                do! this.Response.WriteAsync (sprintf "%s\r\n" strOut)
+            }
+            let writeEndChunk () = task {
+                do! this.Response.WriteAsync ("0\r\n\r\n")
+            }
             if this.Request.Method <> HttpMethods.Head then
-                let serializer = this.GetJsonSerializer()
-                do! serializer.SerializeToStreamAsync dataObj this.Response.Body
+                let pool = ArrayPool<char>.Shared
+                let buffer = pool.Rent 4096
```
TheAngryByrd: Any thoughts on other buffer sizes?

src/Giraffe/ResponseWriters.fs
```diff
@@ -157,9 +160,34 @@ type HttpContext with
         task {
             this.SetContentType "application/json; charset=utf-8"
             this.SetHttpHeader "Transfer-Encoding" "chunked"
+            let writeChunk (strOut : string) = task {
+                // Write the length in hex
+                do! this.Response.WriteAsync (sprintf "%s\r\n" (strOut.Length.ToString("X")))
+                // Write the chunk
+                do! this.Response.WriteAsync (sprintf "%s\r\n" strOut)
```
TheAngryByrd: This could probably be done all in one `sprintf`.

TheAngryByrd:
I should also write some compelling tests for this rather than `curl`ing the SampleApp

src/Giraffe/ResponseWriters.fs
slang25: Looks like a popular default: https://github.com/dotnet/corefx/blob/master/src/System.Text.Json/src/System/Text/Json/Document/JsonDocument.Parse.cs#L15

src/Giraffe/Giraffe.fsproj
```diff
@@ -48,6 +48,7 @@
     <PackageReference Include="Microsoft.AspNetCore.Authentication" Version="2.2.*" />
     <PackageReference Include="Microsoft.AspNetCore.Authorization" Version="2.2.*" />
     <PackageReference Include="Microsoft.AspNetCore.ResponseCaching" Version="2.2.*" />
+    <PackageReference Include="Microsoft.IO.RecyclableMemoryStream" Version="1.2.2" />
```
slang25: I hope the maintenance of this package gets some love, there's been no NuGet releases in over 2 years, but it looks like it could get better: https://github.com/microsoft/Microsoft.IO.RecyclableMemoryStream/issues/62

TheAngryByrd: If this came out of the Bing team I'd _hope_ that they're actively using it, meaning it's been stable.

src/Giraffe/ResponseWriters.fs
TheAngryByrd: I should also probably use `String.Format` instead for lower allocations and speed benefits.

NinoFloris:
Chunking should be automatic in aspnetcore (it emits a chunk per write call), what you have at the moment would be totally incorrect during an http2 connection.

https://github.com/aspnet/AspNetCore/blame/728110ec9ee1b98b2d9c9ff247ba2955d6c05846/src/Servers/Kestrel/test/InMemory.FunctionalTests/ChunkedResponseTests.cs#L19

TheAngryByrd:
If there's a way to defer to AspNetCore's chunking I'm all for it.

TheAngryByrd:
I'm guessing `response.BodyWriter.WriteAsync` is the key, lemme try that out.

TheAngryByrd:
Oh there isn't a BodyWriter until 3.0... hmm.

TheAngryByrd:
Ahah, the main issue is setting the header `this.SetHttpHeader "Transfer-Encoding" "chunked"` causes asp.net to not do the chunked encoding for you.

NinoFloris:
@TheAngryByrd `.Body` is fine

TheAngryByrd:
So I've found another interesting behavior. The chunks that aspnet core writes are `20000` hex or `131072` decimal chunk sizes.  Dotnet core's Stream DefaultBuffer size [is 81920](https://github.com/dotnet/corefx/blob/954b1fdab8e98060cce3d0d88a34c3b0587a6acb/src/Common/src/CoreLib/System/IO/Stream.cs#L31-L34), which is a multiple of 4096 that is below 85000.  However when using this default is seems to write out chunks of `131072`.  The behavior i'm seeing is if it's set to something other than a 4096 * (2^x), it rounds up to the nearest version.


I've set the CopyToAsync buffer to 65536 (or 4096 * 16) which results in it writing `10000` hex or `65536` chunks just to play it safe with Large Object Heap allocations.

NinoFloris:
>So I've found another interesting behavior. The chunks that aspnet core writes are 20000 hex or 131072 decimal chunk sizes. Dotnet core's Stream DefaultBuffer size is 81920, which is a multiple of 4096 that is below 85000.

@TheAngryByrd if those chunks are actually allocated through a single allocation (I think they might be streamed out instead) this is well worth it to open an issue on the aspnetcore repo. Hidden LOH allocations are definitely to be prevented.

TheAngryByrd:
> > So I've found another interesting behavior. The chunks that aspnet core writes are 20000 hex or 131072 decimal chunk sizes. Dotnet core's Stream DefaultBuffer size is 81920, which is a multiple of 4096 that is below 85000.
> 
> @TheAngryByrd if those chunks are actually allocated through a single allocation (I think they might be streamed out instead) this is well worth it to open an issue on the aspnetcore repo. Hidden LOH allocations are definitely to be prevented.

@NinoFloris Yeah I'm unsure if they are actually allocation but I'd rather play it safe.  I was trying to dig and find out where this possibly is happening but I haven't been able to find it.  I'll create a better repro for them and open an issue on the aspnet core repo

src/Giraffe/Serialization.fs
```diff
@@ -14,7 +15,7 @@ module Json =
     open Newtonsoft.Json.Serialization
     open FSharp.Control.Tasks.V2.ContextInsensitive
     open Utf8Json
-
+    let recyclableMemoryStreamManager = RecyclableMemoryStreamManager()
```
slang25: Probably not for this PR, but should we be using this everywhere in place of `MemoryStream`?

TheAngryByrd: I'd say yes unless you know for sure what what you're putting into the Stream is going to be smaller than 85000 bytes. Since it looks like the other places this is used is in other serialization I'd say definitely can be used there.  For the tests it's not really a concern.

TheAngryByrd:
Even though this has to buffer in memory, the addition of `RecyclableMemoryStream` this does fix this issue with large json blobs hitting the Large Object Heap, which is my main concern.

dustinmoris:
@TheAngryByrd  Ok thanks for confirming. Unless @slang25 or @NinoFloris have any objections I'd say let's merge it as it fixes #368.

Issue #370, opened 2019-08-18
Implement routing as a request feature instead of using the Items collection


baronfel:
Was talking with @NinoFloris and he mentioned that currently the routing information/subsystem is implemented using `HttpContext.Items`, which requires unboxing, lookups, etc.  It could be reimplemented using the `Features` concept (with the Items implementation as a fallback?) and sidestep most of those lookups. Implementations of Features that are included in ASP.Net Core can be found [here](https://github.com/aspnet/AspNetCore/tree/4ef204e13b88c0734e0e94a1cc4c0ef05f40849e/src/Http/Http/src/Features).

NinoFloris:
Thanks for adding it :)

slang25:
I had this thought a few months back when I read about `Features`, when I did the original subroutes logic I wasn't aware of it :(
It's a breaking change as custom handlers might rely on the underlying `HttpContext.Items` value, but it's worth doing it sooner than later.

dustinmoris:
Agreed, it would be a welcome breaking change. I might have a stab at this myself!

slang25:
It's also worth noting that Saturn knows about the `Items` `giraffe_route` key and behaviour. So it would make sense to follow up with a PR there, possibly a fallback in the behaviour there, i.e. if the feature is null then revert to the Items lookup.

dustinmoris:
@slang25 Good point, thanks for bringing this to my attention!

dustinmoris:
I'm closing this issue as we agreed in the PR which was actually addressing this issue that it was not the right thing to do. Please feel free to re-open if you disagree!

PR #374, opened 2019-09-07
Feature/sub routing feature

dustinmoris:
Addresses #370 

/cc @baronfel

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -48,6 +48,18 @@ For more information please see [issue #347](https://github.com/giraffe-fsharp/G
 
     type MatchOptions = { IgnoreCase: bool; MatchMode: MatchMode; }
     ```
+- Changed the sub routing feature to use an ASP.NET Core `HttpContext.Features` module for resolving and storing (sub)paths. The previous `SubRouting` module has been deprecated and replaced with the `ISubRoutingFeature` interface and `SubRoutingFeature` class. Additionally there have been added new `HttpContext` extension methods which make it easier to interact with the new sub routing feature:
+
+    - `HttpContext.GetResolvedPath()` replaces `SubRouting.getSavedPartialPath ctx`
+    - `HttpContext.GetNextPartOfPath()` replaces `SubRouting.getNextPartOfPath ctx`
+    - `HttpContext.SetResolvedPath(path)` replaces the manual setting of the `HttpContext.Items` key-value pair
+- The new `ISubRoutingFeature` implementation doesn't use the `HttpContext.Items` collection anymore. For compatibility reasons Giraffe still falls back and populates this key-value pair if run in `Version36` compatibility mode (= default). If compatibility with the `HttpContext.Items` collection is not required then it is recommended to register the Giraffe middleware with `Version40` compatibility:
+
+    ```fsharp
+    let configureApp _ (app : IApplicationBuilder) =
+        app.UseGiraffeErrorHandler(errorHandler)
+            .UseGiraffe(webApp, Version40)
+    ```
 
 ## 3.6.0
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -53,12 +53,12 @@ For more information please see [issue #347](https://github.com/giraffe-fsharp/G
     - `HttpContext.GetResolvedPath()` replaces `SubRouting.getSavedPartialPath ctx`
     - `HttpContext.GetNextPartOfPath()` replaces `SubRouting.getNextPartOfPath ctx`
     - `HttpContext.SetResolvedPath(path)` replaces the manual setting of the `HttpContext.Items` key-value pair
-- The new `ISubRoutingFeature` implementation doesn't use the `HttpContext.Items` collection anymore. For compatibility reasons Giraffe still falls back and populates this key-value pair if run in `Version36` compatibility mode (= default). If compatibility with the `HttpContext.Items` collection is not required then it is recommended to register the Giraffe middleware with `Version40` compatibility:
+- The new `ISubRoutingFeature` implementation doesn't use the `HttpContext.Items` collection anymore. For compatibility reasons Giraffe still falls back and populates this key-value pair if run in `Version36` compatibility mode (= default). If compatibility with the `HttpContext.Items` collection is not required then it is recommended to register the Giraffe services with `Version40` compatibility:
 
     ```fsharp
-    let configureApp _ (app : IApplicationBuilder) =
-        app.UseGiraffeErrorHandler(errorHandler)
-            .UseGiraffe(webApp, Version40)
+    let configureServices (services : IServiceCollection) =
+        services.AddGiraffe(
+            fun cfg -> cfg.CompatibilityMode <- Version40) |> ignore
     ```
 
 ## 3.6.0
--- a/samples/IdentityApp/IdentityApp/Program.fs
+++ b/samples/IdentityApp/IdentityApp/Program.fs
@@ -200,7 +200,7 @@ let configureApp (app : IApplicationBuilder) =
     app.UseCors(configureCors)
        .UseGiraffeErrorHandler(errorHandler)
        .UseAuthentication()
-       .UseGiraffe(webApp, Version40)
+       .UseGiraffe(webApp)
 
 let configureServices (services : IServiceCollection) =
     // Configure InMemory Db for sample application
@@ -242,7 +242,10 @@ let configureServices (services : IServiceCollection) =
     services.AddCors() |> ignore
 
     // Configure Giraffe dependencies
-    services.AddGiraffe() |> ignore
+    services.AddGiraffe(
+        fun options ->
+            options.CompatibilityMode <- Version40
+        ) |> ignore
 
 let configureLogging (builder : ILoggingBuilder) =
     let filter (l : LogLevel) = l.Equals LogLevel.Error
--- a/samples/IdentityApp/IdentityApp/Program.fs
+++ b/samples/IdentityApp/IdentityApp/Program.fs
@@ -200,7 +200,7 @@ let configureApp (app : IApplicationBuilder) =
     app.UseCors(configureCors)
        .UseGiraffeErrorHandler(errorHandler)
        .UseAuthentication()
-       .UseGiraffe webApp
+       .UseGiraffe(webApp, Version40)
 
 let configureServices (services : IServiceCollection) =
     // Configure InMemory Db for sample application
--- a/src/Giraffe/Middleware.fs
+++ b/src/Giraffe/Middleware.fs
@@ -26,26 +26,30 @@ type GiraffeMiddleware (next          : RequestDelegate,
 
     member __.Invoke (ctx : HttpContext) =
         task {
-            let start = System.Diagnostics.Stopwatch.GetTimestamp();
+            let subRoutingFeature = SubRoutingFeature()
+            ctx.Features.Set<ISubRoutingFeature>(subRoutingFeature)
+            try
+                let start = Diagnostics.Stopwatch.GetTimestamp()
+                let! result = func ctx
+                let  logger = loggerFactory.CreateLogger<GiraffeMiddleware>()
 
-            let! result = func ctx
-            let  logger = loggerFactory.CreateLogger<GiraffeMiddleware>()
+                if logger.IsEnabled LogLevel.Debug then
+                    let freq = double Diagnostics.Stopwatch.Frequency
+                    let stop = Diagnostics.Stopwatch.GetTimestamp()
+                    let elapsedMs = (double (stop - start)) * 1000.0 / freq
 
-            if logger.IsEnabled LogLevel.Debug then
-                let freq = double System.Diagnostics.Stopwatch.Frequency
-                let stop = System.Diagnostics.Stopwatch.GetTimestamp()
-                let elapsedMs = (double (stop - start)) * 1000.0 / freq
+                    logger.LogDebug(
+                        "Giraffe returned {SomeNoneResult} for {HttpProtocol} {HttpMethod} at {Path} in {ElapsedMs}",
+                        (if result.IsSome then "Some" else "None"),
+                        ctx.Request.Protocol,
+                        ctx.Request.Method,
+                        ctx.Request.Path.ToString(),
+                        elapsedMs)
 
-                logger.LogDebug(
-                    "Giraffe returned {SomeNoneResult} for {HttpProtocol} {HttpMethod} at {Path} in {ElapsedMs}",
-                    (if result.IsSome then "Some" else "None"),
-                    ctx.Request.Protocol,
-                    ctx.Request.Method,
-                    ctx.Request.Path.ToString(),
-                    elapsedMs)
-
-            if (result.IsNone) then
-                return! next.Invoke ctx
+                if (result.IsNone) then
+                    return! next.Invoke ctx
+            finally
+                ctx.Features.Set<ISubRoutingFeature>(Unchecked.defaultof<ISubRoutingFeature>)
         }
 
 // ---------------------------
--- a/src/Giraffe/Middleware.fs
+++ b/src/Giraffe/Middleware.fs
@@ -15,10 +15,10 @@ open Giraffe.Serialization
 // Default middleware
 // ---------------------------
 
-type GiraffeMiddleware (next                : RequestDelegate,
-                        handler             : HttpHandler,
-                        compatibilityMode   : GiraffeCompatibilityMode,
-                        loggerFactory       : ILoggerFactory) =
+type GiraffeMiddleware (next            : RequestDelegate,
+                        handler         : HttpHandler,
+                        options         : IGiraffeOptions,
+                        loggerFactory   : ILoggerFactory) =
 
     do if isNull next then raise (ArgumentNullException("next"))
 
@@ -27,7 +27,7 @@ type GiraffeMiddleware (next                : RequestDelegate,
 
     member __.Invoke (ctx : HttpContext) =
         task {
-            let subRoutingFeature = SubRoutingFeature compatibilityMode
+            let subRoutingFeature = SubRoutingFeature options.CompatibilityMode
             ctx.Features.Set<ISubRoutingFeature>(subRoutingFeature)
             try
                 let start = Diagnostics.Stopwatch.GetTimestamp()
@@ -97,26 +97,7 @@ type IApplicationBuilder with
     /// Returns `unit`.
     ///
     member this.UseGiraffe (handler : HttpHandler) =
-        this.UseMiddleware<GiraffeMiddleware> (handler, Version36)
-        |> ignore
-
-    /// **Description**
-    ///
-    /// Adds the `GiraffeMiddleware` into the ASP.NET Core pipeline. Any web request which doesn't get handled by a surrounding middleware can be picked up by the Giraffe `HttpHandler` pipeline.
-    ///
-    /// It is generally recommended to add the `GiraffeMiddleware` after the error handling-, static file- and any authentiation middleware.
-    ///
-    /// **Parameters**
-    ///
-    /// - `handler`: The Giraffe `HttpHandler` pipeline. The handler can be anything from a single handler to an entire web application which has been composed from many smaller handlers.
-    /// - `compatibilityMode`: Defines the compatibility mode of the Giraffe routing handlers. It is recommended to use `Version40` if backwards compatibility with the older `SubRouting` module is not required.
-    ///
-    /// **Output**
-    ///
-    /// Returns `unit`.
-    ///
-    member this.UseGiraffe (handler : HttpHandler, compatibilityMode : GiraffeCompatibilityMode) =
-        this.UseMiddleware<GiraffeMiddleware> (handler, compatibilityMode)
+        this.UseMiddleware<GiraffeMiddleware> handler
         |> ignore
 
     /// **Description**
@@ -149,4 +130,14 @@ type IServiceCollection with
         this.TryAddSingleton<IJsonSerializer>(NewtonsoftJsonSerializer(NewtonsoftJsonSerializer.DefaultSettings))
         this.TryAddSingleton<IXmlSerializer>(DefaultXmlSerializer(DefaultXmlSerializer.DefaultSettings))
         this.TryAddSingleton<INegotiationConfig, DefaultNegotiationConfig>()
+        this.TryAddTransient<IGiraffeOptions, GiraffeOptions>()
+        this
+
+    member this.AddGiraffe (configureOptions) =
+        let giraffeOptions = GiraffeOptions() :> IGiraffeOptions
+        configureOptions giraffeOptions
+        this.TryAddSingleton<IJsonSerializer>(NewtonsoftJsonSerializer(NewtonsoftJsonSerializer.DefaultSettings))
+        this.TryAddSingleton<IXmlSerializer>(DefaultXmlSerializer(DefaultXmlSerializer.DefaultSettings))
+        this.TryAddSingleton<INegotiationConfig, DefaultNegotiationConfig>()
+        this.TryAddTransient<IGiraffeOptions>(fun _ -> giraffeOptions)
         this
--- a/src/Giraffe/Middleware.fs
+++ b/src/Giraffe/Middleware.fs
@@ -17,7 +17,7 @@ open Giraffe.Serialization
 
 type GiraffeMiddleware (next            : RequestDelegate,
                         handler         : HttpHandler,
-                        options         : IGiraffeOptions,
+                        options         : GiraffeOptions,
                         loggerFactory   : ILoggerFactory) =
 
     do if isNull next then raise (ArgumentNullException("next"))
@@ -130,14 +130,14 @@ type IServiceCollection with
         this.TryAddSingleton<IJsonSerializer>(NewtonsoftJsonSerializer(NewtonsoftJsonSerializer.DefaultSettings))
         this.TryAddSingleton<IXmlSerializer>(DefaultXmlSerializer(DefaultXmlSerializer.DefaultSettings))
         this.TryAddSingleton<INegotiationConfig, DefaultNegotiationConfig>()
-        this.TryAddTransient<IGiraffeOptions, GiraffeOptions>()
+        this.TryAddTransient<GiraffeOptions, GiraffeOptions>()
         this
 
     member this.AddGiraffe (configureOptions) =
-        let giraffeOptions = GiraffeOptions() :> IGiraffeOptions
+        let giraffeOptions  = GiraffeOptions()
         configureOptions giraffeOptions
         this.TryAddSingleton<IJsonSerializer>(NewtonsoftJsonSerializer(NewtonsoftJsonSerializer.DefaultSettings))
         this.TryAddSingleton<IXmlSerializer>(DefaultXmlSerializer(DefaultXmlSerializer.DefaultSettings))
         this.TryAddSingleton<INegotiationConfig, DefaultNegotiationConfig>()
-        this.TryAddTransient<IGiraffeOptions>(fun _ -> giraffeOptions)
+        this.TryAddTransient<GiraffeOptions>(fun _ -> giraffeOptions)
         this
--- a/src/Giraffe/Middleware.fs
+++ b/src/Giraffe/Middleware.fs
@@ -15,9 +15,10 @@ open Giraffe.Serialization
 // Default middleware
 // ---------------------------
 
-type GiraffeMiddleware (next          : RequestDelegate,
-                        handler       : HttpHandler,
-                        loggerFactory : ILoggerFactory) =
+type GiraffeMiddleware (next                : RequestDelegate,
+                        handler             : HttpHandler,
+                        compatibilityMode   : GiraffeCompatibilityMode,
+                        loggerFactory       : ILoggerFactory) =
 
     do if isNull next then raise (ArgumentNullException("next"))
 
@@ -26,7 +27,7 @@ type GiraffeMiddleware (next          : RequestDelegate,
 
     member __.Invoke (ctx : HttpContext) =
         task {
-            let subRoutingFeature = SubRoutingFeature()
+            let subRoutingFeature = SubRoutingFeature compatibilityMode
             ctx.Features.Set<ISubRoutingFeature>(subRoutingFeature)
             try
                 let start = Diagnostics.Stopwatch.GetTimestamp()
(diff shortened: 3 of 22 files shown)


src/Giraffe/Routing.fs
```diff
@@ -10,37 +10,89 @@ open FSharp.Control.Tasks.V2.ContextInsensitive
...
+        | None   -> this.GetResolvedPathFromItems()
+
+    /// **Description**
+    ///
+    /// Returns the next part of the request path, which hasn't been resolved by a routing handler yet.
+    ///
+    member this.GetNextPartOfPath() =
+        let requestPath = this.Request.Path.Value
+        match this.GetResolvedPath() with
+        // Should this be a StartsWith?
```
dustinmoris: Shouldn't `requestPath.Contains` actually be `requestPath.StartsWith` here?

src/Giraffe/Routing.fs
```diff
@@ -10,37 +10,89 @@ open FSharp.Control.Tasks.V2.ContextInsensitive
...
+        | _ -> requestPath
+
+    /// **Description**
+    ///
+    /// Memorises a partially resolved path between different routing handlers by making use of the SubRoutingFeature.
+    ///
+    member this.SetResolvedPath (path : string) =
+        let subRoutingFeature = this.Features.Get<ISubRoutingFeature>()
+        subRoutingFeature.SetResolvedPath path
+        // Keep this as a fallback for Saturn, etc?
```
dustinmoris: Not sure if this is necessary for Saturn, feels counterintuitive to still populate the Items collection when using a feature now?

src/Giraffe/Routing.fs
```diff
@@ -10,37 +10,89 @@ open FSharp.Control.Tasks.V2.ContextInsensitive
...
+        this.Items.[RouteKey] <- path
+
+    /// **Description**
+    ///
+    /// Clears the currently saved path in the SubRoutingFeature.
+    ///
+    member this.ClearResolvedPath () =
+        let subRoutingFeature = this.Features.Get<ISubRoutingFeature>()
+        subRoutingFeature.ClearResolvedPath()
+        // Keep this as a fallback for Saturn, etc?
```
dustinmoris: Same here...

dustinmoris:
Would like someone else to code review this before mergning my own PR. Thank you!

src/Giraffe/Routing.fs
NinoFloris: I'll do a check, we may only interface with the routing apis, if not I'll submit a pr. If we do so because of an api gap in Giraffe we'll need a compatibility release so we can switch :)

slang25: Here's the code that knows about the items key:
https://github.com/SaturnFramework/Saturn/blob/d4506da6721a592b4601a27f40eae870e7e59e15/src/Saturn/Common.fs#L35-L131

src/Giraffe/Routing.fs
slang25: looking at the usage inside the match, I'd say yes for correctness. Looks like the contains was added to avoid an issue where the request path was changed completely to the error path by the error handling middleware: https://github.com/giraffe-fsharp/Giraffe/commit/a2538f45c4ffab4c5ee263b3658328d93f1cb766

src/Giraffe/Routing.fs
slang25: Could be over engineering here, but we could adopt a similar approach that MVC did, and have a compatibility setting that is defaulted to let's say `3.6`, and consumers opt-in to setting it to `4.0` when they add Giraffe in the startup, then here we'd conditionally do the compatibility thing based on that?

src/Giraffe/Routing.fs
dustinmoris: Coolio, I've changed it to StartsWith now.

src/Giraffe/Routing.fs
dustinmoris: I've added a new `GiraffeCompatibilityMode` type which can be set when registerting the GiraffeMiddleware. I've also started to change existing sub route tests to run in both modes.  I think this is an elegant solution and it's actually a very small change. The most part is just slightly re-factoring existing tests, but that makes testing for this change easier too, so overall a good thing!

src/Giraffe/Middleware.fs
```diff
@@ -96,7 +97,26 @@ type IApplicationBuilder with
     /// Returns `unit`.
     ///
     member this.UseGiraffe (handler : HttpHandler) =
-        this.UseMiddleware<GiraffeMiddleware> handler
+        this.UseMiddleware<GiraffeMiddleware> (handler, Version36)
```
dustinmoris: The default is set to `Version36` if no compatibility mode is set.

src/Giraffe/Routing.fs
```diff
@@ -45,7 +49,10 @@ type HttpContext with
         let subRoutingFeature = this.Features.Get<ISubRoutingFeature>()
         match subRoutingFeature.GetResolvedPath() with
         | Some p -> Some p
-        | None   -> this.GetResolvedPathFromItems()
+        | None   ->
+            match subRoutingFeature.CompatibilityMode with
+            | Version36 -> this.GetResolvedPathFromItems()
+            | Version40 -> None
```
dustinmoris: Only reading from the `Items` collection in compat mode 3.6 now.

src/Giraffe/Routing.fs
```diff
@@ -65,8 +71,9 @@ type HttpContext with
     member this.SetResolvedPath (path : string) =
         let subRoutingFeature = this.Features.Get<ISubRoutingFeature>()
         subRoutingFeature.SetResolvedPath path
-        // Keep this as a fallback for Saturn, etc?
-        this.Items.[RouteKey] <- path
+
+        if subRoutingFeature.CompatibilityMode = Version36 then
+            this.Items.[RouteKey] <- path
```
dustinmoris: Same here, only populating the `Items` key-value pair if running in `Version36` mode.

src/Giraffe/Middleware.fs
slang25: Very nice 😄 
My only thought is that it could be that we change `UseGiraffe` to return an `IGiraffeOptionsBuilder` and then have an extension method `SetCompatibilityVersion`.

The risk is in its current state is that over time we gather up more parameters and it gets a little messy.

Or we could wrap this in a class (not a record as they don't version well).

dustinmoris: Fair point

dustinmoris: I slightly tend more towards the `GiraffeOptions` version, becaues then the `UseGiraffe` extension method doesn't change the return type and it will continue to work in a chain of middleware constelations for existing apps and it will save me to create yet another extension method on the `IGiraffeOptionsBuilder` to `Build()` the middleware and then return an `IApplicationsBuilder`. Any thoughts/tips/reasons why the `IGiraffeOptionsBuilder` might still be better?

slang25: That's a good point, I was just straight up copying MVC 😆 
I don't think you'd expose a `Build` method, you'd do it like `IMvcBuilder`, which has a `Services` property to get back to a fluent API. So the breaking change would be:
```
service.UseGiraffe(app).UseSomeOtherThing();
```
to:
```
service.UseGiraffe(app).Services.UseSomeOtherThing();
```
which is consistent with MVC, Identity Server Authentication, HttpClientFactory and others. With `GiraffeOptions` I haven't given much thought to how it would be to consume.

dustinmoris: I'm just going through the MVC code to see how they've done it. Can you maybe give me some direction where to find the relevant code?

These are the questions which I'm trying to work out:

- The object which will hold the options for Giraffe, is it supposed to get registered as a dependency via the services collection? I suppose that would be the correct place as it might be used beyond just the middleware, right?
- When passing settings through an `IApplicationBuilder` extension method like `SetCompatibilityMode(...)` can I then register that object in the `IServiceCollection` too or would I need to set it a second time in `AddGiraffe()`?

samples/IdentityApp/IdentityApp/Program.fs
```diff
@@ -242,7 +242,10 @@ let configureServices (services : IServiceCollection) =
     services.AddCors() |> ignore
 
     // Configure Giraffe dependencies
-    services.AddGiraffe() |> ignore
+    services.AddGiraffe(
+        fun options ->
+            options.CompatibilityMode <- Version40
+        ) |> ignore
```
dustinmoris: @slang25 Is this what you had in mind? I think this is certainly more future proof and ASP.NET Core-ish, but I wonder if for Giraffe it wouldn't be better to have a `Func<IGiraffeOptions, IGiraffeOptions>` as a parameter intead of an `Action<IGiraffeOptions>`? It would be more functional to accept an object of `IGiraffeOptions` and then return a modified version of this object as part of the configuration instead of making direct modifications on the original `IGiraffeOptions` parameter. What do you think?

src/Giraffe/Middleware.fs
dustinmoris: Ok, I've made another change, see below for how it works.

samples/IdentityApp/IdentityApp/Program.fs
dustinmoris: e.g.:

```fsharp
services.AddGiraffe(fun options -> options.SetCompatibilityMode(Version40))
```

The `.SetCompatibilityMode` method would then return a new object with the mode being changed. This would also allow chaining of multiple settings if this object is going to grow in the future, e.g.:

```fsharp
services.AddGiraffe(
    fun options -> 
        options
            .SetCompatibilityMode(Version40)
            .SetSomethingElse(123)
            .SetYetAnotherValue(true)
)
```

slang25: I'm away from my laptop to have a proper look, but this looks nice. Will have a proper look tonight.

src/Giraffe/Middleware.fs
```diff
@@ -125,4 +130,14 @@ type IServiceCollection with
         this.TryAddSingleton<IJsonSerializer>(NewtonsoftJsonSerializer(NewtonsoftJsonSerializer.DefaultSettings))
         this.TryAddSingleton<IXmlSerializer>(DefaultXmlSerializer(DefaultXmlSerializer.DefaultSettings))
         this.TryAddSingleton<INegotiationConfig, DefaultNegotiationConfig>()
+        this.TryAddTransient<IGiraffeOptions, GiraffeOptions>()
+        this
+
+    member this.AddGiraffe (configureOptions) =
```
slang25: How about optional:
```suggestion
    member this.AddGiraffe (?configureOptions) =
```

NinoFloris: This really stings if you use it from C# aspnet core host projects, if we're going with Func and Action etc this should be Optional; DefaultValue(null)

NinoFloris: Or just add an overload

samples/IdentityApp/IdentityApp/Program.fs
NinoFloris: Also my PR (even though it got brutishly hijacked) got in for dotnet core 5 https://github.com/aspnet/Extensions/issues/1463 so we can actually make IOptions usable for records by then.

NinoFloris:
So if this isn't in too big of a rush what I'd like to do is submit a PR to Saturn to remove its reliance on the Item, as it could really just use routing functions that are exposed by Giraffe (primarily `getSavedPartialPath`). Would that be preferable to building in compatibility switches or should we do it side by side?

samples/IdentityApp/IdentityApp/Program.fs
slang25: I don't think that helps us here (it's a great change though), we aren't touching `IOptions` here.
I have generally found that records in public interface for things like options is great until you add a field, then every consuming `with` usage is broken, so any additions would be a major version change.

src/Giraffe/Middleware.fs
slang25: Just so I understand, is the fact it's not optional the bit you object to, or is the optional fsharp parameter suggested above still problematic?

samples/IdentityApp/IdentityApp/Program.fs
dustinmoris: @slang25 That's an interesting point about the use of record types!

dustinmoris:
@NinoFloris That would be nice, but I don't want to couple a Giraffe release to a Saturn release. Having to coordinate releases is always a bit of a pain. I don't mind the compatibility switch. It's not a huge change and sooner or later this might be required for other features too, so at least there's already an extendable template in place.

src/Giraffe/Middleware.fs
dustinmoris: Currently the `member this.AddGiraffe (configureOptions)` is an overload. The original `member this.AddGiraffe ()` is still there. Did you guys want me to reduce it to one method with an optional parameter?

slang25: Nope, that was my bad, what you have here is better 🙂

dustinmoris: Ok thank you :)

dustinmoris:
Ok chaps, I'll update the documentation then and unless there's any other feedback I'd like to merge this if that's ok?

RELEASE_NOTES.md
```diff
@@ -48,6 +48,18 @@ For more information please see [issue #347](https://github.com/giraffe-fsharp/G
...
+
+    - `HttpContext.GetResolvedPath()` replaces `SubRouting.getSavedPartialPath ctx`
+    - `HttpContext.GetNextPartOfPath()` replaces `SubRouting.getNextPartOfPath ctx`
+    - `HttpContext.SetResolvedPath(path)` replaces the manual setting of the `HttpContext.Items` key-value pair
+- The new `ISubRoutingFeature` implementation doesn't use the `HttpContext.Items` collection anymore. For compatibility reasons Giraffe still falls back and populates this key-value pair if run in `Version36` compatibility mode (= default). If compatibility with the `HttpContext.Items` collection is not required then it is recommended to register the Giraffe middleware with `Version40` compatibility:
+
+    ```fsharp
+    let configureApp _ (app : IApplicationBuilder) =
+        app.UseGiraffeErrorHandler(errorHandler)
+            .UseGiraffe(webApp, Version40)
```
slang25: This needs updating with the recent changes

samples/IdentityApp/IdentityApp/Program.fs
NinoFloris: @slang25 wasn't claiming it was, just good to know of its existence. It's definitely useful for classes too, as it pertains to any type with non empty constructors.

src/Giraffe/Middleware.fs
NinoFloris: Yep overload is much better. If you would need an optional value that borders interop with C# libs it's usually nicer to do `[<Optional;DefaultParameterValue(null)>]` instead of `?param` that way it'll show up as a standard optional value we all know from C#.

slang25: Oh nice, didn't know that!

src/Giraffe/Routing.fs
```diff
@@ -10,37 +10,97 @@ open FSharp.Control.Tasks.V2.ContextInsensitive
...
-// Private sub route helper functions
+// Sub Routing Feature
 // ---------------------------
 
-[<RequireQualifiedAccess>]
-module SubRouting =
+[<Literal>]
+let private RouteKey = "giraffe_route"
+
+type ISubRoutingFeature =
```
NinoFloris: Is this interface going to be enough? Afaik F# isn't going to support DIM implementations for a while, so versioning it could become nasty.

src/Giraffe/Routing.fs
```diff
@@ -10,37 +10,97 @@ open FSharp.Control.Tasks.V2.ContextInsensitive
...
+    /// **Description**
+    ///
+    /// Returns a partially (or fully) resolved request path during routing decisions when working with the SubRoutingFeature.
+    ///
+    member this.GetResolvedPath() =
+        let subRoutingFeature = this.Features.Get<ISubRoutingFeature>()
+        match subRoutingFeature.GetResolvedPath() with
+        | Some p -> Some p
+        | None   ->
+            match subRoutingFeature.CompatibilityMode with
```
NinoFloris: I think this may be the only primitive method that should really exist as an extension on HttpContext, as it's actually used by Giraffe users. The other three not so much (GetNextPartOfPath, SetResolvedPath, ClearResolvedPath) and would be better exposed through a single method into the feature, say `GetSubRouting: unit -> ISubRoutingFeature` .

src/Giraffe/Routing.fs
```diff
@@ -10,37 +10,97 @@ open FSharp.Control.Tasks.V2.ContextInsensitive
...
 
-    let getNextPartOfPath (ctx : HttpContext) =
-        match getSavedPartialPath ctx with
-        | Some p when ctx.Request.Path.Value.Contains p -> ctx.Request.Path.Value.[p.Length..]
-        | _   -> ctx.Request.Path.Value
+    /// **Description**
+    ///
+    /// Returns a partially (or fully) resolved request path during routing decisions when working with the SubRoutingFeature.
+    ///
+    member this.GetResolvedPath() =
```
NinoFloris: GetSubRoute?

src/Giraffe/Routing.fs
```diff
@@ -10,37 +10,97 @@ open FSharp.Control.Tasks.V2.ContextInsensitive
...
+        | None   ->
+            match subRoutingFeature.CompatibilityMode with
+            | Version36 -> this.GetResolvedPathFromItems()
+            | Version40 -> None
+
+    /// **Description**
+    ///
+    /// Returns the next part of the request path, which hasn't been resolved by a routing handler yet.
+    ///
+    member this.GetNextPartOfPath() =
```
NinoFloris: While it's an extension method it actually contains critical routing logic (how to get from resolved path to next part). Wouldn't it be wise to put this in the interface too?

src/Giraffe/Routing.fs
NinoFloris: Continuing, isn't this interface so specific that there's not a lot of value in it? If so you could argue that you could better represent this as a single property on a concrete class and be done with it.

src/Giraffe/Routing.fs
```diff
@@ -10,37 +10,97 @@ open FSharp.Control.Tasks.V2.ContextInsensitive
...
+let private RouteKey = "giraffe_route"
+
+type ISubRoutingFeature =
+    abstract member CompatibilityMode : GiraffeCompatibilityMode with get
+
+    abstract member GetResolvedPath   : unit   -> string option
+    abstract member SetResolvedPath   : string -> unit
+    abstract member ClearResolvedPath : unit   -> unit
+
+type SubRoutingFeature (compatibilityMode : GiraffeCompatibilityMode) =
```
NinoFloris: If this implementation has the compatibility mode, why not have the middleware that creates this pass in HttpContext too? Then you can have these implementations handle setting the old Items based key.

src/Giraffe/Routing.fs
```diff
@@ -10,37 +10,97 @@ open FSharp.Control.Tasks.V2.ContextInsensitive
...
+        subRoutingFeature.SetResolvedPath path
+
+        if subRoutingFeature.CompatibilityMode = Version36 then
+            this.Items.[RouteKey] <- path
+
+    /// **Description**
+    ///
+    /// Clears the currently saved path in the SubRoutingFeature.
+    ///
+    member this.ClearResolvedPath () =
```
NinoFloris: Path -> Route

src/Giraffe/Routing.fs
```diff
@@ -10,37 +10,97 @@ open FSharp.Control.Tasks.V2.ContextInsensitive
...
+        let requestPath = this.Request.Path.Value
+        match this.GetResolvedPath() with
+        | Some p when requestPath.StartsWith p -> requestPath.[p.Length..]
+        | _ -> requestPath
+
+    /// **Description**
+    ///
+    /// Memorises a partially resolved path between different routing handlers by making use of the SubRoutingFeature.
+    ///
+    member this.SetResolvedPath (path : string) =
```
NinoFloris: Path -> Route

src/Giraffe/Routing.fs
NinoFloris: Method naming is good here.

dustinmoris:
@NinoFloris I've made most changes as requested. Some of it led to a massive simplification of the interface, which exposes only two methods now, which is really what the rest of Giraffe is interested in.

gerardtoconnor:
> Pinging @gerardtoconnor for his take on routing.

Hey guys, had quick look through, great work.

If I were to be overly critical, and feel free to ignore as just being purist, but what are we hoping to achieve here? I notice we are now doing all kinds of compatibility checks now at runtime which I don't like for maintainability or performance … and I am not sure what we are gaining here?

For reference, I may be wrong but the intention appears to be "reduce boxing" from using items dictionary … just so we're all on same page, there is a big difference in both performance and jit operation between boxing a value type and casting a reference type. The IFeature instance, like items dictionary, is retrieved via a cast on an object dictionary (IFeature Collection), so every time we pull it out we are casting. We have to hold the subpath in the IFeature anyway so all we seem to be doing is creating and allocating an extra object each request and sort of doing the reverse of what we intended of "perf". If we consider the string we currently use in Items dictionary, string is a reference type so we are casting to and from object, not boxing, with casting being a lot cheaper but still involving JIT checks. I can quite plausibly see this being slower then the original implementation given that subroutes are often only shallow/few, with obj -> string cast being a lot cheaper then transient DI instance allocations of IFeature.

IFeature is classic case of (poor  IMHO) OOP design, abstraction instead of data/structure

I think it would be very important to do an __accurate verified perf test on this before merging it__ because unless there is many subroute mutations (which is rare), I would envisage this performing poorer then our current setup, all while adding the compatibility mess. Hopefully I am wrong and it's faster but when complexity and compatibility are increasing, always measure twice and cut once?

Feel free to disregard, I was against Items dictionary originally also but that was why I built token router which takes different approach, I think building optimised routing structures for both routing and meta/swagger is a better way forward but appreciate everyone else's input.

NinoFloris:
@gerardtoconnor it's actually very simple, `Items` is backed by a feature too so it incurs at least twice the lookup cost. At least feature lookup is a highly optimized type map because `Items` is indeed a slow dictionary.

Combined with the changes @dustinmoris did in https://github.com/giraffe-fsharp/Giraffe/pull/374/commits/f0fc9e28eafccc0dc71c00403435ce777bf7e81f the feature is now also doing most work on its own instance field, what would have been repeated `Items` lookups before. 
Leaving the downside of this PR to be the extra compatibility branch. That could be resolved by having two implementations of the interface but most likely it is entirely insignificant because of branch prediction operating on what is essentially a constant.

Though one necessary change I do see is that `SubroutingFeature` should have the options injected by the middleware that creates it instead of retrieving them from `RequestServices` per method call, basically like it was before. @dustinmoris 

https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.features.itemsfeature

dustinmoris:
Chaps, I've taken both your feedback on board and pushed yet another change. I think this probably the nicer solution but keen to get more input from you!

gerardtoconnor:
@NinoFloris Items may look like a feature (and is) but is treated differently as for every HttpContext default initialisation gets this injected [into a struct of properties to avoid lookup on standard Features](https://github.com/aspnet/HttpAbstractions/blob/bc7092a32b1943c7f17439e419d3f66cd94ce9bd/src/Microsoft.AspNetCore.Http/DefaultHttpContext.cs#L213).

As i mentioned on my comment, appreciate its likely more costly over many many calls, vs amortising one IFeature cast, but I'm not a fan of the requirement now to include and check compatibility constraints at runtime and do prefer alternative method/code path for these sorts of things.

We are still substringing loads and Saturn's/Giraffe's parsing is suboptimal so was just cautious of introducing new and weird OOP abstractions and questioning if the extra complexity was worth it hence the recommendation of an extensive perf test before merging.

Another Issue I foresee here is that I estimate this will be fairly immaterial while now introducing a compatibility API that will need to be catered for going forward, but ideally we want to move to a (more compatible & flexible) token router to facilitate Analysers & Swagger going forward, as well as speed in lookup and parsing, but if it is a material improvement in benchmarks, it is worth the introduction & merge of the compatibility complexity 👍 .

NinoFloris:
>default initialisation gets this injected into a struct of properties to avoid lookup on standard Features.

Nice didn't know features was part of that ensemble too!

> but I'm not a fan of the requirement now to include and check compatibility constraints at runtime and do prefer alternative method/code path for these sorts of things.

Agreed though it's at least zero cost with the last commit, it's indeed a maintenance burden and one I didn't mind solving by removing the unnecessary dependency on the Item key in Saturn so we wouldn't have to introduce this at all.

>weird OOP abstractions 

Can you explain why? Is an interface bad, or is `Items` better in your opinion?

> Another Issue I foresee here is that I estimate this will be fairly immaterial

I think so too, I don't expect this to move the needle much compared to compiler optimized tasks for instance.

I think something like endpoint routing (or token router) will definitely be the way forward if only because you get introspection. But I don't quite remember the reasons why we didn't default to token router in the past, I was only a spectator in that discussion.

slang25:
I think that token routing is the way to go in order to get OpenAPI introspection, we won't get it as a result of Endpoint Routing directly, and token routing using Endpoint Routing, but it's an orthogonal thing in my mind.

NinoFloris:
>but it's an orthogonal thing in my mind.
Well not entirely of course, you need one to have any shot at providing the other. At this point we have zero introspection options besides parsing source code.

slang25:
Good point, I just contradicted myself I the same comment

gerardtoconnor:
> At least feature lookup is a highly optimized type map because Items is indeed a slow dictionary.

Looking over code again, every call to get/set feature values is going through the IFeature collection call with cast, the only speed up I can see is the lookup speed of IFeature type dictionary vs Items hash dictionary … and the difference will be immaterial in the scheme of things. The allocation of a IFeature each request will surely offset this in high throughput situations?

Apart from the fact that I think the perf will be the same if not slower … I am not quite sure why we are doing this compat split, is it only due to Saturn needing & using Items dict? Surely it can use old dependency version until its updated to use the new IFeature if that is the case or is there other reasons?

> weird OOP abstraction
> Can you explain why? Is an interface bad, or is Items better in your opinion?

Purely opinion but deterministic > abstraction (except for testing), It's bad IMO as we are slathering in OOP abstraction FAT when all our state is is a simple string ... I prefer to work on data then abstract into black boxes, especially when the black box allocates in my high throughput server (for testing substitution is needed so abstraction is well suited). It's probably not a debate for here 😉 .

> But I don't quite remember the reasons why we didn't default to token router in the past, 

Token router needed to route then process pipeline so was little less flexible then current method, endpoint routing doesn't help us on parsing and StringFormat routes so it could be used but only partially ... would need to tested with routef parsing to see how would work ... but some kind of structure is my preferred option also per @slang25 comment.

I'm guessing this is too far along to not merge to develop but I would stress once again that I would do that only to perf test it vs master ... as per all our original performance PRs we merged, we tested the performance thoroughly before accepting to master.

dustinmoris:
> Surely it can use old dependency version until its updated to use the new IFeature if that is the case or is there other reasons?

True

> I'm guessing this is too far along to not merge to develop but I would stress once again that I would do that only to perf test it vs master ... as per all our original performance PRs we merged, we tested the performance thoroughly before accepting to master.

I'm not precious about this PR. Code is just code, so if something gets binned again then it's not going to affect me personally :)

I am holding back the merge for now. This is a good discussion and I looked at the code which Gerard linked above and I'm quite impressed how they optimised the access to the IFeatureCollection.Items property via a cached ref on the hot path. Definitely learned something in the last 20 minutes digging through the ASP.NET Core implementation there :)

gerardtoconnor:
> Pinging @gerardtoconnor for his take on routing.

Narrator:_"@NinoFloris regretted this ping"_ 😂 

>  and I'm quite impressed how they optimised the access to the IFeatureCollection.Items property

It would have been criminal if this optimization was not in place if it is standard interface for HttpContext, like if we had a Dictionary behind the IFeature methods to grab a string!?!

> I'm not precious about this PR. Code is just code, so if something gets binned again then it's not going to affect me personally :)

I am sometime sentimental about code so I wouldn't presume 😄 .

Sorry to railroad things on this PR, was not my intention and think benchmarking it still worth it … the main issue I have is from a maintainability standpoint with this compatibility API, and would rather avoid it if we can... as well as extra complexity.

If we want to improve performance, the obvious roads IMO are:

1. Improved async perf via `ply` or something else
2. Improved routing & parsing via look up tree structure like token router, possibly using endpoint routing

NinoFloris:
> Surely it can use old dependency version until its updated to use the new IFeature if that is the case or is there other reasons?
> True

This would depend on the version this change would be in, the version bound for Saturn referencing Giraffe is >= 3.4.0 && < 4.0.0 according to nuget, so merging this into a minor release would break things.

> every call to get/set feature values is going through the IFeature collection call with cast, the only speed up I can see is the lookup speed of IFeature type dictionary vs Items hash dictionary … and the difference will be immaterial in the scheme of things

Agreed and with the news that `Items` doesn't go through Features I'm not expecting this is a performance improvement beyond noise. I do think it's a nicer API to have this route string in a feature (as I suggested in the PR at some point) which is very similar to [IEndpointFeature](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.features.iendpointfeature). But leaving it in `Items` and not having to make any change is very justified to keep questionable churn to a minimum.

> Pinging @gerardtoconnor for his take on routing.
>Narrator:"@NinoFloris regretted this ping" 😂

Not at all, it's precisely why I pinged you ;)

gerardtoconnor:
> This would depend on the version this change would be in, the version bound for Saturn referencing Giraffe is >= 3.4.0 && < 4.0.0 according to nuget, so merging this into a minor release would break things.

On a side note, as this is a breaking change, if it were to be integrated, it needs to be done on the next major release (4.0.0) for this reason. Think it was already agreed long ago that we were following semantic versioning so this would violate this with a breaking minor change. In statically typed complied languages there is no place for runtime compatibility checking/handling for packages/functionality on the same platform/runtime.

NinoFloris:
As the code is right now it defaults to the 34 compatibility level, so not sure what the actual break is in that case. However if you'd do what you suggested without a major bump that'd break hard so I didn't think that was a better plan than what we already had.

gerardtoconnor:
@NinoFloris yeah i see that it defaults to 3.4 compat but then it once again becomes question of why are we even doing this!? Only a handful of people would know about/use this feature, even if it was faster … and it feels like patchwork fix to problem that doesn't exist. 

I guess the point I am trying to get across is that going forward, for maintainability and simplicity, all breaking changes (that are heavily tested for both function and perf) are postponed into major releases, and compatibility work arounds should be avoided if at all possible. Just my opinion of course, but I have only experienced pain in the past when trying to build in runtime compatibility branching etc.

> However if you'd do what you suggested without a major bump that'd break hard so I didn't think that was a better plan than what we already had.

The recommendations for async/ply improvement, as well as route tree are definitely breaking changes for version 4.0.0, agree completely.

NinoFloris:
I think we all individually came to the conclusion that it doesn't feel like it's worth the effort. So no need to convince anyone I guess ;)

gerardtoconnor:
Ah I'm not trying to unsell this PR any further don't worry, just planting the seed against any further compatibility hacks for the future 😉

slang25:
I'm at the same conclusion too, it's not the right thing to do. As tidy as the implementation is!

NinoFloris:
@dustinmoris time to close this PR don't you think?

dustinmoris:
@NinoFloris Yes, I think we all agreed that this won't get merged and that there is no point in changing anything because the `Items` implementation is not any worse than this.

PR #379, opened 2019-09-30
Endpoint Routing

dustinmoris:
Basic concept of getting Giraffe with endpoint routing working.

The idea is that we create a new module called `EndpointRouting` which will basically expose an extension method called something like `MapGiraffeRouter` for `IRouteBuilder` (this would be similar to MVC's `MapControllers`).

The `MapGiraffeRouter` will take in some object/list/function (not sure what it will exactly look like) of a router. This router will let users basically register one top level router where paths are mapped to `HttpHandler` functions.

Then the `MapGiraffeRouter` will iterate through all these routes and create a RequestHandler for each mapping and essentially map the Giraffe HttpHandler functions to routes which are registered in the ASP.NET Core Endpoint Routing middleware.

I've created a simple proof of concept in this PR which would let a user pass in a list of simple path strings tupled with `HttpHandler` functions. This would be sort of enough if Giraffe would only have the `route` handler. Obviously there is more work to do for things like `routef` etc, but the concept should still work.

Ideas, feedback, counter suggestions, everything welcome!

EDIT:

My thinking is that this will essentially defer the entire routing logic, performance optimisatinos, etc. to the endpoint routing middleware. It will also therefore nicely interop with other middleware which already works today with endpoint routing and even more in the future.

This feature will make the `GiraffeMiddleware` redundant, as the routing will be handled by the endpoit routing middleware and we just register Giraffe endpoints during Startup via `.UseEndpoints(c => c.MapGiraffeRouter())`.

Because it doesn't require the `GiraffeMiddleware` we can initially release this as a side by side feature. People can either `UseGiraffe` in ASP.NET Core with existing stuff, or `UseEndpoints` with the new router logic.


dustinmoris:
Removed everything except `netcoreapp3.0` for simplicity whilst we work on this PR. We can still add the other frameworks back in again in the end if we don't want to release this as part of a major release.

dustinmoris:
Also this explains endpoint routing pretty well now: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/routing?view=aspnetcore-3.0

src/Giraffe/EndpointRouting.fs
```diff
@@ -0,0 +1,33 @@
...
+            return ()
+        } :> Task
+
+type IRouteBuilder with
+    member private this.MapGiraffeGetRoute (path : string, func : HttpFunc) =
+        let requestDelegate = convertToRequestDelegate func
+        this.MapGet(path, requestDelegate)
+        |> ignore
+
+    member this.MapGiraffe (routes : (string * HttpHandler) list) (handler : HttpHandler) =
```
slang25: What's the trailing handler for?

dustinmoris: What, which trailing handler are you speaking of? *quicklyremovingandpushingnewchange*

It's not required, just a result of me typing before thinking.

slang25:
This is good, and demonstrates the touch points with endpoint routing. But it leaves out the hardest part, what do those routes look like and how do we map them, currently it's a list of paths and handlers, and they are assumed to be GETs.

Do we start with the token routing which looks like this:
```fsharp
let notFound = setStatusCode 404 >=> text "Not found"
let app =
    router notFound [
        route "/"       (text "index")
        route "/about"  (text "about")
    ]
```
and extract it from the trie structure, or do we build up a different structure. Is our route constraints compatible with the endpoint routes constraints, or will we need to reconsider how `routef` functions?
Or do we take a totally different approach?

gerardtoconnor:
I think the basic route methods are non-trivial enough, the complexity comes when we try to use `routef` methods, endpoint router seems to use named keys, sting constraints, and extracted values are stings that need to then be converted to the actual type value (from extracted string). The token router's tree is probably overly complex for intermediate tree representation, easier tree for route/subroutes should suffice for non-parse URLs, `routef` will be the tough part to do efficiently. May need to abandon PrintF StingFormat type safety to use fully which is a pity and big compromise.

dustinmoris:
Thanks for both your input. I realised that my initial example was oversimplistic. I've slightly expanded on my initial example to get `routef` pretty much fully working (was too lazy to do the Guid mapping, but otherwise should work as expected). I have also added a new sample app which demonstrates the current proof of concept and can be tested.

> But it leaves out the hardest part, which is what do those routes look like and how do we map them, currently it's a list of paths and handlers, and they are assumed to be GETs.

Yes, I've currently assumed them as GETs, but how we want to map the HTTP verb is the smallest problem. We could add another parameter into the list of `path`, `delegate` tuples which includes the verb and then we know which `Map[...]` method to use, or we can invent slightly new handlers (e.g. `get`, `getf`, `post`, `postf`, etc.), or probably do something else really clever somewhere.

I don't know what you mean by how do we map them, so I've added a sample application. You can pull this PR and run the new sample app and see it working in action. The mapping is really just done by the mapping methods like `MapGet` or `MapSomething`.

> and extract it from the trie structure, or do we build up a different structure

I would argue a new strucute, maybe as simple as a stupid list in this current PoC. The reason why is because the trie structure is giving us all the benefit of being able to resolve an incoming request and map to the correct handler in a very efficient way. However, this part is now deferred to the routing middleware (which probably uses a trie structure itself I'd guess) and therefore we don't need such a complex structure anymore for the simple task of registering handlers with the routing middleware. A list might just be sufficient.

> Is our route constraints compatible with the endpoint routes constraints, or will we need to reconsider how routef functions?

Fair question. I hope my latest commit answers this one. Our routing methods are all compatible with the new endpoint routing constraints, because to be fair, the endpoint routing constraints allow even more flexible and complex rules than what we have, so what we have can be already easily mapped with the constraints. Before we had to convert a format string into Regex, now we almost use the same logic to convert the format string into another format string. I didn't have to do anything different (as in new) from what we had before, just a slightly different conversion, but no more or less work.

> the complexity comes when we try to use routef methods, endpoint router seems to use named keys, sting constraints, and extracted values are stings that need to then be converted to the actual type value (from extracted string).

As shown in the example it's actually not difficult at all. It's also not any different from what we did with Regex before. The Regex match also extracted the found values as strings into groups and we had to parse them into the actual types.

> and extracted values are stings that need to then be converted to the actual type value

Actually, the routing middleware already does that for us for all primitive types. I can remove the parsing for all our types except hte special Guid types, so there is an optimisation in my current sample code which I haven't done yet.

> he token router's tree is probably overly complex for intermediate tree representation

Agreed

> routef will be the tough part to do efficiently.

Got it working now :), but I have not spent any thoughts to optimise yet, however I think it's not a bad place to start with. The conversion from `HttpHandler` to `RequestDelegate` and the conversion from the `PrintF` string to the template string  happens once during app startup, then both is being registered with the middleware. That is actually already a huge improvement, because the original `routef` has to do the conversion into regex every time when a request comes in.

Anyways, these are just my thoughts based on my latest example, but I'm waiting for your feedback and feel free to rip things apart, I really enjoy your input an all those things!

src/Giraffe/EndpointRouting.fs
```diff
@@ -0,0 +1,134 @@
...
+open Microsoft.AspNetCore.Routing
+open Microsoft.FSharp.Reflection
+open FSharp.Core
+open FSharp.Control.Tasks.V2.ContextInsensitive
+
+module private RouteTemplateBuilder =
+    let private getConstraint (i : int) (c : char) =
+        // Route Constraint Docs:
+        // https://docs.microsoft.com/en-us/aspnet/core/fundamentals/routing?view=aspnetcore-3.0#route-constraint-reference
+        (dict [
```
gerardtoconnor: Are we able to just convert this to a match statement instead? this is creating a new dictionary every time its called which is a bit wasteful and rather then lifting it to a static value, since all values are constant / known at compile time, match statements should always be used as match will be faster then dictionary 99% of the time thanks to jump table optimizations and branch locality

src/Giraffe/EndpointRouting.fs
```diff
@@ -0,0 +1,134 @@
...
+            'u', ShortId.toUInt64    >> box  // uint64
+        ]
+
+    let private convertToTuple (formatChars : char list) (routeData : RouteData) =
+        let values =
+            (routeData.Values.Values, formatChars)
+            ||> Seq.map2 (fun v c ->
+                let parser = parsers.[c]
+                let value  = parser (v.ToString())
+                value)
```
dustinmoris: I think this can actually be optimised. We could change `parsers.[c]` into something like `getSpecialParser c` which then returns either `None` or `Some parser`. Most things like `i`, `d`, `f`, etc. will immediately return `None` and only the special cases like `u` will return a parser (e.g. `ShortId.toUInt64`).

This means that for most values we can just return `v` instead of doing `let value = parser (v.ToString())`. I think `v` is already the correct type resolved by the endpoint routing middleware based on the constraint which we provided (e.g. {someInt:int}), so there is actually a potential gain over the original `routef` function too.

slang25:
Nice! I'll have a good look in the morning.

src/Giraffe/EndpointRouting.fs
```diff
@@ -0,0 +1,134 @@
...
+            'i', int32               >> box  // int
+            'd', int64               >> box  // int64
+            'f', float               >> box  // float
+            'O', parseGuid           >> box  // Guid
+            'u', ShortId.toUInt64    >> box  // uint64
+        ]
+
+    let private convertToTuple (formatChars : char list) (routeData : RouteData) =
+        let values =
+            (routeData.Values.Values, formatChars)
```
gerardtoconnor: RouteData.Values is a hybrid key/value dictionary, can we be sure that enumerating it will be in the same order as our char list? if it was a hash based one it would likely not be, it is probably stack based one so should be fine but we should test to make sure

src/Giraffe/EndpointRouting.fs
```diff
@@ -0,0 +1,134 @@
...
+            ||> Seq.map2 (fun v c ->
+                let parser = parsers.[c]
+                let value  = parser (v.ToString())
```
gerardtoconnor: v.ToString() feels a little hacky here, thought RouteData dict containted strings but it is Objects so it might be worth inspecting the underlying objects being pulled here as they may be already boxed converted values so we are doing conversion back to string and then once again the value boxed?

gerardtoconnor:
This routef approach is like what I was thinking, annoying they allocate strings/objects for each parse value, instead of creating value types straight from url scan, but that's probably a little pedantic and unavoidable if we're using endpoint routing, worth it for the speedup over regex matching and swagger support. I was hoping the obj might be the boxed parse value but don't thing so as doc state: 

> The framework-provided route constraints don't modify the values stored in route values. All route values parsed from the URL are stored as strings. For example, the float constraint attempts to convert the route value to a float, but the converted value is used only to verify it can be converted to a float.

We will need to ensure composition chains works here as well, hopefully fine.

Only bit left will be to figure out subrouting etc, where a primitive list/tree might be needed but other then that looks good to me! 👍

dustinmoris:
> We will need to ensure composition chains works here as well, hopefully fine.

Absolutely, I think it should work just fine, but will give it a try.

> Only bit left will be to figure out subrouting etc, where a primitive list/tree might be needed but other then that looks good to me! 👍

Agreed. Wow, is this the famous gerardtoconnor stamp of approval? =)

Thanks for all the early feedback! Looks like I'm on the right path with this appraoch then. I'll keep working on this feature in the open now and periodically push my changes to this PR so everyone can see what's going on and intervene, code review and have an influence in the end result of Giraffe's endpoint routing story. Also, everyone is welcome to push changes themselves to this PR!

gerardtoconnor:
>Wow, is this the famous gerardtoconnor stamp of approval? =)

Yikes, sorry, don't mean to come across overly critical... I am trying to relax a bit on efficeincy correctness (at least in OSS), it's nice if you have the time but rolling out functionality adds more value, especially with so many requesting swagger functionality. Thanks for getting POC out there and ball rolling!

dustinmoris:
@gerardtoconnor Ha, I eant it mostly in a jokey way. Your input is always hugely appreciated! Keep pushing for all those nitty gritty perf improvements, because if not you, who else? It's been immensely useful to have different people with different focuses to help make Giraffe better at every corner!

slang25:
I've still yet to kick the tires on this, but let me quickly address your comment @dustinmoris. I pretty much agree with everything you've said, and poking about the token routing stuff have also realised that it is not to be used as a useful representation beyond an optimised structure for routing.

> I don't know what you mean by how do we map them

Yeah, I wasn't clear. The way I think of a Giraffe app is that you can put `choose` anywhere in the hierarchy, and now you've branched, that could be potentially a new route. So you have a tree structure, how do we flatten and map that to routes? I think for simplicity it makes sense for what you have done, that is have a top level structure which is flat, but you could still branch off inside a handler deeper inside.

This leads me to a question I'm still wondering, what do we want to solve here? As I understand this is about providing a way to better fit in with the rest of ASP.NET Core.
If we also want OpenAPI information, it would be handy if this new structure let's us introspect it and get all the information we need for that, as I don't see well get it as a direct result of using Endpoint Routing.

In terms of the current implementation, and wiring up `routef`, very nice 😃. A thought I had, I think Endpoint Routing allows you to create custom `RouteConstraints`, would it be possible for us to implement the Giraffe ones, rather than needing to map to the vanilla route format? I'll do some investigation.

dustinmoris:
> Yeah, I wasn't clear. The way I think of a Giraffe app is that you can put choose anywhere in the hierarchy, and now you've branched, that could be potentially a new route. So you have a tree structure, how do we flatten and map that to routes? I think for simplicity it makes sense for what you have done, that is have a top level structure which is flat, but you could still branch off inside a handler deeper inside.

Correct. The Giraffe routing was originally inherited from the model which Suave had in place. Then it got slightly improved where it does things slightly differently than Suave's original routing. With Endpoint Routing we should probably ask ourselves the critical question if something more than a flat top level strucutre is even required (in 99% of applications)? Routing on a path and HTTP verb are the main routing mechanisms in most applications. I feel like they could be all satisfied with a flat top level routing list. Other subsequent routing decisions (e.g. if a model was posted in JSON or XML, or by inspecting the Accept header, etc.) can then still happen via a choose in the attached handler, which wouldn't affect the endpoint registered route anyway.

Maybe that's an opoprtunity to re-think some current handlers..

> This leads me to a question I'm still wondering, what do we want to solve here? As I understand this is about providing a way to better fit in with the rest of ASP.NET Core.

Absolutely, it is to me clear that ASP.NET Core will push for endpoint routing across the board. In the future if someone says that the CORS middleware doesn't work correctly with MVC, then the MS response will be: Use Endpoint Routing. If the ASP.NET Core team is going to develop new authentication functionality, they will most and foremost make it work really nicely with the registered endpoints and only half hearted retrofit it into pre-endpoint routing MVC routes, etc.

There's no proof for this to happen, but I'm convinced and for that reason alone I want Giraffe to be part of that journey, with a great endpoint routing API which will eventually become the default routing API in templates, etc.

Secondly, I expect the ASP.NET Core team to predominantely improve the endpoint routing performance if they do perf work on routing for future releases, and less for the old routing logic in MVC and Razor Pages, etc. So in this case I want Giraffe to be on that train too.

> If we also want OpenAPI information, it would be handy if this new structure let's us introspect it and get all the information we need for that, as I don't see well get it as a direct result of using Endpoint Routing.

For me OpenAPI was and still isn't a Giraffe user story. MVC doesn't have their own OpenAPI implementation either. Nancy doesn't have it, etc. It just happens that MVC is so widely used that someone else put in the effort to build OpenAPI into MVC, which is Swashbuckle. Ideally I would have openend an issue in Swashbuckle and said "As a Giraffe user I want Swashbuckle to generate a Swagger document for Giraffe apps too", but the developers would have probably said "Not on our priority list, but feel free to submit a PR".

With endpoint routing there will be lots of other (non Giraffe) users who will go in and say "I want Swashbuckle to read API information from endpoint routing, so that routes which I've registered from non MVC middleware get also picked up" and they are more likely to put in the work themselves and make that happen now and everyone, including Giraffe users get OpenAPI for free.

At least it's not unreasonable to tell Giraffe users now: If you want OpenAPI then go to the OpenAPI repo and open there a ticket taht it doesn't work with ASP.NET Core's default new routing, which is endpoint routing.

There is also a chance that MS themselves will (or maybe already secrelty is) working on an OpenAPI story based on their new routing API.

> A thought I had, I think Endpoint Routing allows you to create custom RouteConstraints, would it be possible for us to implement the Giraffe ones, rather than needing to map to the vanilla route format?

Sounds interesting! Where can I read up about 'RouteConstraint'?

dustinmoris:
> At least it's not unreasonable to tell Giraffe users now: If you want OpenAPI then go to the OpenAPI repo and open there a ticket taht it doesn't work with ASP.NET Core's default new routing, which is endpoint routing.

Just to stress one thing that was always on my radar when thinking of Giraffe. I always wanted Giraffe to be built in such a way that it integrates with the entire .NET OSS and MS system. If someone makes a cool ASP.NET Core auth layer then I want Giraffe to be just working with it, because we designed auth handlers to use the same underlying ASP.NET Core APIs. Same for OpenAPI. I don't want users to be in a situation where they struggle to get OpenAPI work with Giraffe, because the way we do routing meant that it was difficult to integrate with existing open source packages which already do that work. It made us as a bottleneck, and that is the worst. I believe the reason why Giraffe has been adopted so quickly is because it mostly let people do what they wanted to do with it and integrate it with almost anything else. Routing was an exception so far, but endpoint routing is IMHO the end of that bottleneck. As soon as Giraffe has a good endpoint routing API, then this should become the standard in my opinion and make Giraffe even more open and compatible with everything else. This is the thing which I didn't like about Suave in the beginning, that it has its own version of everything which evidently makes innovation and feature development slow, because the maintainers are the bottleneck, and there's only so much time a handful of humans have alongside their 9-5 work, kids, wives, friends, etc.

baronfel:
One thing to keep in mind here is how Saturn uses routing right now. It lands heavily on the ability to insert a subroute-level choose block at any point in the patching, and it sounds like this is being relegated to the backburner here.

gerardtoconnor:
@slang25 
>poking about the token routing stuff have also realised that it is not to be used as a useful representation beyond an optimised structure for routing.

It's a specifically specialized tree that includes state transitions of matching fixed and variable tokens into leaf functions, it has no other uses but to match & parse out urls quickly as not a data structure abstraction ... I will benchmark latest Token router vs Endpoint for throughput at some stage but would be confident token router still faster thanks to its direct value type parse application while for basic routes Endpoint would be faster thanks to intrinsics.

@dustinmoris 
>  With Endpoint Routing we should probably ask ourselves the critical question if something more than a flat top level structure is even required (in 99% of applications)? Routing on a path and HTTP verb are the main routing mechanisms in most applications. 

This is (ironically) one of the original reasons Token Router was not integrated by default, the need to finalise routes up front with no/little child routing capabilities further in the handler chain. Like you are pointing out, it is not an issue for most users but may have knock on effects for Saturn, Saturn probably being the biggest user. How are we proposing handing subroutes / subcontrollers as it stands?

On endpoint / routing speed, would note that the majority of urls in LOBA will be parsed and as parsing is magnitudes slower then matching strings/bytes it will dilute the benefits of intrinsics on a holistic level (unless you can utilise to parse directly into your function). So the main benefits I see here are alignment with aspnet core, improved fuzzy matching over regex, and handing off OpenAPI considerations to Endpoint team ... I think after addressing Saturn concerns, next areas for optimisation are the parsing / function application, seeing if boxing and temp structures can be reduced.

On `RouteConstraints`, you have already started to integrated these, and for further constraints like on expected character length/max size etc, this could be mapped legally using standardised StringFormat width constaints eg `%8i` , `+` / `-` symbols are allowed, so can be used to convey less than/greater than requirements eg `%-5i` / `%+7i`. Adding in specialised handlers for Endpoint's regex and other non-routef cases is also probably good easy win leveraging the constraints api. [Further reading on constraints here](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/routing?view=aspnetcore-3.0#route-constraint-reference)

slang25:
Noooooooo, too many console windows open and force pushed the wrong thing!!! Trying to fix now

slang25:
Shit  😞 @dustinmoris are you able to `git push -f` commit b90113d  🙏

I have `git add -A;git commit --amend --no-edit;git push -f` in my shell history and ran it in the wrong shell tab whilst working on another repo. What's worse is that I don't have commit b90113d locally to force push back.

slang25:
Update: ok managed to fix this in the end. GitHub keeps the commits in the case of a force push *phew*.

Nothing to see here...

ssiltanen:
Hope this PR progresses soon as this is very much anticipated feature 👍

Krzysztof-Cieslak:
Hey. 

What’s the status for this PR? It completely terrifies me and I have no idea how I’d support it in Saturn without breaking all the stuff... but I guess I’d want mentally prepare in case you plan to merge/release it ;-)

dustinmoris:
@Krzysztof-Cieslak 

I'm planning to make a couple more changes in this PR (over this weekend maybe) and then merge it into the main branch under a new module/namespace so it lives side by side to the existing suff without breaking anything. It will be an early opt-in feature which I will release as v5.0.0-alpha-001 or something, so users who want to start using Giraffe with Endpiont routing can pull a new prerelease package.

Not related to endpoint routing, but as part of that alpha release I'll be also removing the GiraffeViewEngine from Giraffe and expose this as a new NuGet package called `Giraffe.ViewEngine`, this will benefit those users who want to use the hot reload feature with a smaller binary to speed up local web dev.

Back to the endpoint routing, as it is an opt in feature initially it will require the user to open a new namespace and then the routing handlers will be overridden with the new endpoint functionality. If Saturn is using the Giraffe routes under the hood then I think it should *just work*, but I'm happy to help you and work out how that would affect Saturn and any PRs or suggestions are always welcome, that's why its going to be pre-release to begin with!

Krzysztof-Cieslak:
Oh, I do realize it won't be breaking change but opt-in feature (and will probably need to do the same in Saturn)... but at the same time it seems Endpoint routing is a way to go in longer-term - new stuff from MSFT is using it (for example this reverse-proxy thing) so the community will also move in this direction and will expect to have that in Saturn.  

The part of the problem is that Saturn design utilizes heavily that everything, both routing and behaviour, was the same `HttpHandler` while Endpoint routing keeps those 2 things separated. 
We also utilize the it was possible to create hierarchical routing structures (using subrouting to ensure that components just declare relative routing and then it can be composed together)  with "old" Giraffe routing, while Endpoint routing kinda requests us to have a flat list in the end.  

For example, let's say we have the following `controller`:
```
let myController = controller {
   index SomeIndexHttpHandler
   add SomeAddHttpHandler
   update SomeUpdateHttpHandler

   version "1"
   plug [All] (PlugAllHttpHandler)
   plug [Add] (PlugAddHttpHandler)
```

Under the hood, this would be transformed to the following `HttpHandler` (more or less, this is a simplified version) 

```
checkVersion "1" >=> choose [
    GET >=> choose [
       route "/add" >=> PlugAllHttpHandler >=> PlugAddHttpHandler >=> SomeAddHttpHandler
       route "/" >=> PlugAllHttpHandler >=> SomeIndexHttpHandler ]
    POST >=> choose [
       routef "%id" (fun id -> PlugAllHttpHandler >=>  (SomeUpdateHttpHandler id)
     ]
```

Then you can have parent `router`, that's routing requests into this `controller` - and `router`  

```
let apiRouter = router {
   pipe_through SomeOtherHttpHandler

   get "/health" myHealthHandler
   forward "/books" myController
}
```

Which would be:

```
SomeOtherHttpHandler >=> choose [
   GET >=> route "/health" >=> myHealthHandler
   subRoute "/books" >=> myController
]
```

And then I could put this `router` into another `router` etc - everything was `HttpHandler` so people could compose this decision tree really easily. 
So as you can see the whole design is about fluently mixing routing and behaviours because everything was, in the end, the same abstraction - `HttpHandler`

I think general concepts are portable to Endpoint routing - it will just require quite a sizeable changes - things like `router` or `controller` won't be transformed down to `HttpHandler` but to list of `Endpoint`, and then functions like `get` or `forward` will require to have overloads - you either plug `Endpoint` or `HttpHandler`. It will break some scenarios (for example manually composing `myHttpHandler >=> myController` ) but I hope we can live with that.

TBF, I quite like the whole concept of endpoint routing - such separation of routing is why we have `route` CE in Saturn - it's just our current solution was really specialized and created for how "old" Giraffe routing worked, and this will be different. But it's fine, everyone needs to adapt.

Krzysztof-Cieslak:
> Not related to endpoint routing, but as part of that alpha release I'll be also removing the GiraffeViewEngine from Giraffe and expose this as a new NuGet package called Giraffe.ViewEngine, this will benefit those users who want to use the hot reload feature with a smaller binary to speed up local web dev.

I like that change, especially that `Feliz.ViewEngine` exists now, so users will be able to switch more easily if they'd prefer different DSL.

0x53A:
Not directly related to Endpoint Routing, but related to Swagger/OpenAPI, which seems to be a major motivation for this:

Do you have any plans for how to add additional metadata to an endpoint?

With MVC/Swashbuckle it automatically picks up XML Comments for the methods and parameters, and there are a bunch of Swashbuckle Attributes to add additional info like [``SwaggerResponse``](https://github.com/domaindrivendev/Swashbuckle.AspNetCore#enrich-response-metadata).

Krzysztof-Cieslak:
@dustinmoris, can I somehow help with finishing this PR?

dustinmoris:
@Krzysztof-Cieslak Yes, want to chat via a Google Hangout or some other tool to quickly brief you in?

slang25:
I'm happy to do the same, would be good to catch up with you both and am curious to hear thoughts on this PR

TheAngryByrd:
Live Stream! 😀

dustinmoris:
Ha sounds good. I'm happy to live stream this, could actually chat this....

Wait let me check my calendar...

Today... lockdown, hmm
Tomorrow... lockdown.. aha
Friday, Saturday, Sunday, bloody hell... looks like my calendar is fully
greyed out with lockdown for the forseeable future..

Ok I think I can do this evening, or tomorrow evenig, or the evening after
that, or <pick your day> 😁

Any suggestions?

Krzysztof-Cieslak:
Yeah, happy to set up stream - I've already done streams with a bunch of guests so I should have everything figured out. I have some other plans today, but tomorrow afternoon/evening would work for me.

dustinmoris:
@Krzysztof-Cieslak Speak tomorrow then!

Krzysztof-Cieslak:
I've rebased this on the latest `develop`, let's see if CI is still passing and if so it will be ready to merge as we've discussed last week.

Krzysztof-Cieslak:
Yay, CI green. Anyone want to review it so we can merge and release 5.0-alpha? ;-)

Giraffe.sln
```diff
@@ -37,6 +37,10 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "_build", "_build", "{8F5B79
 EndProject
 Project("{6EC3EE1D-3C4E-46DD-8F32-0CC8E7565705}") = "Giraffe.Benchmarks", "tests\Giraffe.Benchmarks\Giraffe.Benchmarks.fsproj", "{2188E634-B828-4629-89B9-3680422460F5}"
 EndProject
+Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "EndpointRoutingApp", "samples\EndpointRoutingApp\EndpointRoutingApp\EndpointRoutingApp.fsproj", "{E4E1FB62-8268-4E0C-9F6E-25FF21E376E9}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "EndpointRoutingApp", "EndpointRoutingApp", "{C8D24EFD-B9A9-43DD-BE58-BB3C673869A5}"
```
slang25, line 42: For some reason, Rider isn't happy about the way the nested projects are defined now. If I remove the endpoint routing solution dir and project and re-add it it's happy again 🤷‍♂️

samples/EndpointRoutingApp/EndpointRoutingApp/Program.fs
```diff
@@ -0,0 +1,63 @@
+﻿open Microsoft.AspNetCore
```
slang25, line 1: I'd like to see some unit tests covering the new endpoint routing code, I guess starting with the examples here. Happy to pick that up after this PR 🙂

slang25:
Nice one! 🤘

PR #389, opened 2019-12-04
Move some dependences to not be used on netcoreapp3.0.

isaacabraham:
This PR will help generate significantly smaller dependencies for clients with Paket that use netcore3.


slang25:
Closed with #396 
Thanks for this!

PR #372, opened 2019-08-20
Adding ToIsoString method and unit tests

MorganW09:
New ToIsoString methods to resolve issue #367 around wrong date string being produced.

PR #393, opened 2019-12-17
Updating documentation to align with previous code change.

MorganW09:
Updating documentation to align with change I introduced several months ago in #372. 

I know code PRs should be opened against dev but since this is just a documentation change I thought master would be fine? Let me know if not and I can open against the dev branch.


slang25:
Thanks for this @MorganW09

MorganW09:
Glad I could contribute again!!! Thanks for merging it in!

PR #398, opened 2020-02-09
Reduced Dependencies

dustinmoris:
I'm proposing to release a new version of Giraffe with these fewer dependencies which are not required for netcore3.0 anymore and which cause some Paket issues for users according to Isaac Abraham who maintains the SAFE stack.

Any objections?

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -1025,7 +1025,7 @@ Please be aware that such a `routex` can create a conflict and unexpected behavi
 let webApp =
     choose [
         routex "/foo(/*)" >=> text "Bar"
-        routef "/foo/%s/%s/%s" (fun s1, s2, s3 -> text (sprintf "%s%s%s" s1 s2 s3))
+        routef "/foo/%s/%s/%s" (fun (s1, s2, s3) -> text (sprintf "%s%s%s" s1 s2 s3))
 
         // If none of the routes matched then return a 404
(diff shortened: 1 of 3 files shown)


slang25:
I talked to Isaac about this, my feeling are we should release these changes, but it won't really have much effect, the biggest gain to be had is getting a release version of `TaskBuilder.fs`, see here:
 https://github.com/rspeele/TaskBuilder.fs/issues/29

forki:
@dustinmoris no objections. Please release and when @rspeele releases the new task builder then the whole ecosystem can release a new wave.

isaacabraham:
@slang25 actually, these changes have a bigger impact in terms of e.g. paket lock files than the task builder (which at least is already available as an alpha).

slang25:
@isaacabraham Ah, I realise what's going on here.

You are absolutely right, Paket does the right thing here, the problem is if you are not using paket, dotnet restore (nuget) will pick `NETStandard.Library` `1.6.1` rather than `2.0.3`, and it's a complete mess.

slang25:
Perhaps Giraffe should explicitly lift `NET.Standard` as it is a known transitive dependency. I feel a PR coming.

slang25:
PR created here, not sure it's the right thing though #400

isaacabraham:
@slang25 @dustinmoris any chance this could get released? it would really help with the getting started experience.

PR #400, opened 2020-02-10
Raise NETStandard.Library version to 2.0.3

slang25:
Relates to discussion in #398

With the sample app, the `project.assets.json` file (a sort of transient `paket.lock` for nuget) drops from `4830` lines vs `1300` lines.

A downside to this is:
- It's weird to have to do this
- We get a compiler warning in the Giraffe project that is hard to suppress

I'm not convinced we should merge this, I think the real solution is to press on with a `2.2.0` `TaskBuilder.fs`, or migrate to `ply`.

### DIFF
--- a/src/Giraffe/Giraffe.fsproj
+++ b/src/Giraffe/Giraffe.fsproj
@@ -56,6 +56,11 @@
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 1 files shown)


dustinmoris:
Agree this feels like an unnecessary hack. TaskBuilder already has the change in an alpha release so we should just push for the proper release there or we could release a pre-release of Giraffe with the pre-release of TaskBuilder for those who can't wait.

slang25:
Agreed 👍

PR #412, opened 2020-04-24
Move regex's out of hot-path

dbrattli:
Move the regexp out of the hot-path so it will only be evaluated once when pre-compiling the handler pipeline, and not for every request


forki:
Awesome

PR #414, opened 2020-05-01
Added JsonOnlyNegotiationConfig convenience INegotiationConfig implementation

mcon:
# Summary
It's quite a common usecase to have an API server that only returns JSON, and not XML: any server which wants to behave in this way currently requires a boilerplate `INegotiationConfig` implementation - this PR adds that implementation to Giraffe for re-use.

# Testing
 - [x] Existing HttpHandlerTests passing
 - [x] Ensure new JsonOnlyNegotiationConfig has corresponding testcases in HttpHandlerTests (will probably apply `[<MemberData>]` to existing testcases.

# Follow up
The existing tests in `HttpHandlerTests.fs` could be much more concise - I'm happy to refactor these in a subsequent PR.


dustinmoris:
Thanks for the PR, looks great! I've changed the target to be the `develop` branch, which currently is in preparation for a new major release with some breaking changes which I plan to release initially as a pre-release version. If you need this change faster as part of a stable release then I can hotfix `master`, but I'm reluctant if there's no reason so please let me know!

> The existing tests in HttpHandlerTests.fs could be much more concise - I'm happy to refactor these in a subsequent PR.

That would be fantastic! Really appreciate the help!

PR #418, opened 2020-05-22
Add System.Text.Json Support

slang25:
- Updated SDK to 3.1.300
- Switched to just `netcoreapp3.1`
- Removed mono references from build pwsh scripts
- Added System.Text.Json serializer


TheAngryByrd:
Looks good to me. I think we just need additions to the DOCUMENTATION.md Json section.

slang25:
Docs updated, I think this is good for review now 😃

Issue #48, opened 2017-04-27
File upload


nojaf:
It would be nice if we have some helper methods when uploading files.

Based on the [aspnet core docs](https://docs.microsoft.com/en-us/aspnet/core/mvc/models/file-uploads) and this [stackoverflow issue](http://stackoverflow.com/questions/36437282/dealing-with-large-file-uploads-on-asp-net-core-1-0) I came up with [this](https://gist.github.com/nojaf/2b8eb64b4232558dbf5ba238c576bceb) but it the code isn't all that nice.

dustinmoris:
Hi, just to let you know I will look at your example later this evening and pick this one up next!

nojaf:
Thanks!

dustinmoris:
Hi,

I spent some time looking into this and I created an example in the `SampleApp`. I have added two new HTTP POST endpoints, one for [small file uploads and another for large file uploads](https://github.com/dustinmoris/Giraffe/blob/develop/samples/SampleApp/SampleApp/Program.fs#L131-L132):

```fsharp
POST >=>
    choose [
        route "/small-upload" >=> smallFileUploadHandler
        route "/large-upload" >=> largeFileUploadHandler ]
```

Then I implemented [two handlers](https://github.com/dustinmoris/Giraffe/blob/develop/samples/SampleApp/SampleApp/Program.fs#L90-L111) to deal with small and large file uploads:

```fsharp
open System.Threading
open Microsoft.AspNetCore.Http.Features

let smallFileUploadHandler =
    fun (ctx : HttpContext) ->
        async {
            return!
                (match ctx.Request.HasFormContentType with
                | false -> setStatusCode 400 >=> text "Bad request"
                | true  ->
                    ctx.Request.Form.Files
                    |> Seq.fold (fun acc file -> sprintf "%s\n%s" acc file.FileName) ""
                    |> text) ctx
        }

let largeFileUploadHandler =
    fun (ctx : HttpContext) ->
        async {
            let formFeature = ctx.Features.Get<IFormFeature>()
            let! form = formFeature.ReadFormAsync CancellationToken.None |> Async.AwaitTask
            return!
                (form.Files
                |> Seq.fold (fun acc file -> sprintf "%s\n%s" acc file.FileName) ""
                |> text) ctx
        }
```

[...]

PR #419, opened 2020-05-27
Clarify file upload streaming

cmeeren:
Note: Please only merge if this is actually correct.

When I read the Giraffe docs, I didn't understand that you can stream files with the `IFormFeature` method. I thought the docs mentioned two different buffering upload mechanisms, and then referred to the oh-my-gosh-that's-a-lot-of-boilerplate streaming code shown in the linked MS docs (which do not mention `IFormFeature`) for streaming. But I then came across https://github.com/giraffe-fsharp/Giraffe/issues/48#issuecomment-302924969, and my understanding is now that the `IFormFeature` method is a low-boilerplate way to stream files.

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -2087,7 +2087,7 @@ let fileUploadHandler =
 let webApp = route "/upload" >=> fileUploadHandler
 ```
 
-For large file uploads it is recommended to [stream the file](https://docs.microsoft.com/en-us/aspnet/core/mvc/models/file-uploads#uploading-large-files-with-streaming) in order to prevent resource exhaustion.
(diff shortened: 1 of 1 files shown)


cmeeren:
It seems that I was wrong; [`IFormFeature` buffers all the files](https://stackoverflow.com/a/62217960/2978652). This makes me wonder why @dustinmoris in https://github.com/giraffe-fsharp/Giraffe/issues/48#issuecomment-302924969 called `IFormFeature` handler a "large file handler"; perhaps a misunderstanding?

dustinmoris:
You're right, the naming in the comment is wrong. I checked that the documentation has it right though: https://github.com/giraffe-fsharp/Giraffe/blob/master/DOCUMENTATION.md#file-uploads

Thanks for looking into this!

Issue #416, opened 2020-05-11
Form binding of composite/complex types


christianfolie:
In an application we are sending form data via HTTP-POST. It's part of a bigger, structured model and we have roughly the following behaviour:

- we have a basic model containing properties of another type (see example below)
- within the basic model some of these properties are `option`s
- data within the form is 'composed' with property name prefixes
- we are sending the form data with a plain HTML `form.submit` and try to do a model binding via `TryBindFormAsync<CompositeModel>`
- The binding fails with errors like `Missing value for required property Name.` regardless if it's used with or without the `option`

As binding on arrays/sequences with sub-types work I was expecting to use sub types as well, but it seems it's not implemented yet.
Searching in the issues for 'Model binding' and 'form' revealed no issues, so I don't know if this is missing intentionally or if I am missing something.

ATM I am drafting a failing unit tests and playing a little bit with the implementation and to see if it's possible.
The failing sample so far:

```
[<CLIMutable>]
type Child =
{
    Name : string
    Age  : int
}

[<CLIMutable>]
type CompositeModel = {
    FirstChild: Child
    SecondChild: Child option
}
```

Failing tests (will move them into a PR if desired)
```
[<Fact>]
let ``ModelParser.parse with composite model and SecondChild missing data`` () =
    let modelData =
        dict [
            "FirstChild.Name"  , StringValues "FirstName"
            "FirstChild.Age"   , StringValues "2"
        ]
    let expected =
        {
            FirstChild = { Name = "FirstName"; Age = 2}
            SecondChild = None
        }
    let culture = None
    let result = ModelParser.parse<CompositeModel> culture modelData
    Assert.Equal(expected, result)
    
[<Fact>]
let ``ModelParser.parse with complete composite model data`` () =
    let modelData =
        dict [
            "FirstChild.Name"  , StringValues "FirstName"
            "FirstChild.Age"   , StringValues "2"
            "SecondChild.Name" , StringValues "SecondName"
            "SecondChild.Age"  , StringValues "10"
        ]
    let expected =
        {
            FirstChild = { Name = "FirstName"; Age = 2}
            SecondChild = Some { Name = "SecondName"; Age = 10}
        }
    let culture = None
    let result = ModelParser.parse<CompositeModel> culture modelData
    Assert.Equal(expected, result)
```

dustinmoris:
[...]

PR #422, opened 2020-05-28
Composite model bindings should work

christianfolie:
As mentioned in [this issue](https://github.com/giraffe-fsharp/Giraffe/issues/416#issuecomment-635001722) a PR with some tests for binding complex types.

I'm working on a version to solve this binding, but ATM it's only working in some cases, I might push it later.


dustinmoris:
@christianfolie Thanks for submitting these tests. I've just implemented the required changes in order to make them pass and therefore implement this feature!

PR #423, opened 2020-05-28
Feature/defer routing delegate creation

Krzysztof-Cieslak:
This defers creation of RequestDelegate to the last moment before plugging into middleware. Thanks to keeping `HttpHandler` in the `Endpoint` type we will be able to combine HttpHandlers even after we created routing for it. I've hit this issue in Saturn - in there we have the possibility to plug additional `HttpHandler` for every request coming to the application, and it's defined in a different place (on `application` level) then we define routing (`router`). 

Also, I can imagine this will be useful for other use cases, and may enable more fluent mixing of Endpoints and HttpHandlers - I can imagine having something like [pseudocode] `additional_http_handler >+> Endpoint` which will get translated to `Endpoint(additional_http_handler >=> old_handler)`


Krzysztof-Cieslak:
I've added `applyBefore` and `applyAfter` helper functions that provide helpers for the composition of HttpHandlers with Endpoints

Krzysztof-Cieslak:
@dustinmoris or anyone else, any opinions about this? I'd love to get that in next alpha version as it unblocks certain things in Saturn

Krzysztof-Cieslak:
Hey, @dustinmoris are you OK with us merging those PRs (like this or #429)? Or do you want to accept them personally before merging? Because it seems we have small decision paralysis here 😁

dustinmoris:
Yes looks good. I've quickly renamed 'SingleEndpoint' into 'SimpleEndpoint'. Previously it made sense to call is `single` when the other one was `nested`, but now that we also have a `TemplateeEndpoint` I think `SimpleEndpoint` is a better name. What do you think?

PR #429, opened 2020-06-07
Move to xml docs

jchannon:
Thought I'd start work on moving the markdown docs to xml. VSCode can now use this format and Rider only uses this format.

Will do the other files as and when in separate PRs or one big one, whatever you think!


Krzysztof-Cieslak:
As someone who was pushing people towards markdown format - I 👍 this change. I've fucked up pretty badly back in the days, by not considering cross-editor support for generated/suggested documentation format.

baronfel:
Lovely work, thank you for doing this.

NinoFloris:
@jchannon Nice work! 

Big PR is fine :)

src/Giraffe/ComputationExpressions.fs
```diff
@@ -1,30 +1,31 @@
...
     member __.ReturnFrom v = v
     member __.Zero()       = None
 
-/// **Description**
-///
-/// Enables control flow and binding of `Option<'T>` objects
-///
 let opt = OptionBuilder()
 
+/// <summary>
```
jchannon, line 22: moved the docs to the type - seems to make more sense but could be missing something

src/Giraffe/ComputationExpressions.fs
```diff
@@ -1,30 +1,31 @@
...
 /// A collection of F# computation expressions:
 ///
 /// `opt {}`: Enables control flow and binding of Option<'T> objects
 /// `res {}`: Enables control flow and binding of Result<'T, 'TError> objects
-///
+/// </summary>
 module Giraffe.ComputationExpressions
 
+/// <summary>
```
jchannon, line 9: moved the docs to the type - seems to make more sense but could be missing something

Krzysztof-Cieslak:
Can you target this on the `develop` branch rather than `master` ?

NinoFloris:
It seems previous PR targeted master too so we already have divergence on develop and master. @dustinmoris to minimize confusion for contributors we should change default branch to develop, thoughts?

src/Giraffe/FormatExpressions.fs
```diff
@@ -83,20 +83,13 @@ let private convertToRegexPatternAndFormatChars (mode : MatchMode) (formatString
...
-/// **Output**
-///
-/// Matched value as an option of 'T
-///
+/// </summary>
+/// <param name="format">The format string which shall be used for parsing.</param>
+/// <param name="options">The options record with specifications on how the matching should behave.</param>
+/// <param name="input">The input string from which the parsed arguments shall be extracted.</param>
+/// <returns>Matched value as an option of 'T</returns>
 let tryMatchInput (format : PrintfFormat<_,_,_,_, 'T>) (options : MatchOptions) (input : string) =
```
jchannon, line 93: Screenshot shows that it shows the generic args as a bit odd but again might be my noob understanding. It does this in VSCode and Rider. Screenshot 2020-06-07 at 19 56 52

src/Giraffe/Core.fs
```diff
@@ -268,86 +170,68 @@ type HttpContext with
...
 // ---------------------------
 
-/// **Description**
-///
-/// A type alias for `Task<HttpContext option>` which represents the result of a HTTP function (`HttpFunc`).
-///
-/// If the result is `Some HttpContext` then the Giraffe middleware will return the response to the client and end the pipeline. However, if the result is `None` then the Giraffe middleware will continue the ASP.NET Core pipeline by invoking the `next` middleware.
-///
+/// <summary>
+/// A type alias for <see cref="System.Threading.Tasks.Task<HttpContext option>" />  which represents the result of a HTTP function (HttpFunc).
```
jchannon, line 174: VSCode doesn't seem to like generics on xmldocs so doesn't display it

src/Giraffe/FormatExpressions.fs
baronfel: This is what it looks like for me in Ionide/VSCode:

image

That seems reasonable, I'm not quite sure what you're seeing in terms of generic args wierdness? Do you mean the `'a, 'b, 'c, 'd` parameters that aren't explicitly specified in the code but are implicitly provided by the compiler?

jchannon: I mean after my changes in Rider and I assume rider it has `e is a, f is b`

forki:
Awesome work

Jonathan Channon <notifications@github.com> schrieb am Di., 9. Juni 2020,
21:01:

dustinmoris:
Thanks for all the work! Just a quick question, what should I be seeing now when I hover a function? I'm getting the raw XML in Rider for some reason:
image

jchannon:
This is a bug in rider which I have logged to @auduchinok

Issue #269, opened 2018-05-15
Handling Option<'a> types in Json binding


whitebear-gh:
Hello,

I'm not sure if this is something you might be interested in but according to [this blog post](https://dusted.codes/giraffe-110-more-routing-handlers-better-model-binding-and-brand-new-model-validation-api) it's is something you have in roadmap. 

In my project (we still use Giraffe in beta) we wanted to have support for functional types for Models including requiring non-optional to be existing. In example if we have model like following: 
```fsharp
type Model =
{
  stringProp: string
  optionalProp: string option
}
```

you will get an error trying to call an API with object without `stringProp`, but it will be ok if `optionalProp` is not provided and serializer will set it's value as `None`. With that there is no need to additionally check if value is null or not and you get nice message if required property is missing. If this is something you might be interested in here is the code we use: 

[Gist](https://gist.github.com/whitebear-gh/f2a107961c5024ba61d032976292431a)

Regards
wb

baronfel:
Similarly to this I've got a set of converters that do things pretty seamlessly, including some that work for DU cases where the properties are all out of order: https://github.com/baronfel/Newtonsoft.Json.FSharp.Idiomatic

dustinmoris:
Thanks to both of you for sharing your code in order to add this feature. When I get around to implement it then I will get back to you if I have any questions. Alternatively if you feel like you want to submit a PR to help with the `TryBindJson` function then you'd be more than welcome!

dustinmoris:
I think this is a stale issue which has actually been addressed with #433.

PR #433, opened 2020-06-19
Add `TryBindJson` support for Newtonsoft.Json

mcon:
Implements `TryBindJson` for Newtonsoft.Json serializer, as suggested in https://github.com/giraffe-fsharp/Giraffe/issues/269, thanks @whitebear-gh for the gist.

Looks like it should be possible to get this working for UTF8 and `System.Text.Json`.


Krzysztof-Cieslak:
1. Please target the `develop` branch, rather than `master`
2. Giraffe now moved to XML comments from markdown format (#429)

dustinmoris:
This looks great! I'm going to properly review it when this PR gets marked as `Ready for review`, currently it is still in `Draft` so I don't want to start the PR before all changes have been added. I've resolved the merging conflict, so now this PR can be merged against `develop` again.

mcon:
Only just returned to this one (thanks for merging in develop @dustinmoris), looked into the UTF8Json and System.Text.Json cases, and it's not really possible to implement this for those cases in a meaningful way because there's no way to specify that non-option types are mandatory.

It seems the best that can be done for these serializers (at the moment) is to just compare to the default value for a type, which doesn't really cut it (despite being MS's suggested workaround in the System.Text.Json case https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to#required-properties).

There's an outstanding issue to implement this in UTF8Json (https://github.com/neuecc/Utf8Json/issues/70), but no indication it's going to get done, and it's not clear how this could be implemented in an allocation-free way (one of the big UTF8Json selling points).

Given it's probably not desirable to add this functionality for a single serializer I'm going to go ahead and close this one (unless you think otherwise!).

PR #437, opened 2020-06-29
Add metadata for Endpoints

Krzysztof-Cieslak:
Enhances `Endpoint` type with an additional field for storing Metadata (it's just `obj list` - metadata is not typed in ASP .NET). Under the hood transformed into `.WithMetadata` call on the `IEndpointRouteBuilder`.
It also adds `addMetadata` helper that enables users to add Metadata to their endpoint.

### DIFF
--- a/src/Giraffe/EndpointRouting.fs
+++ b/src/Giraffe/EndpointRouting.fs
@@ -179,11 +179,12 @@ type HttpVerb =
 
 type RouteTemplate = string
 type RouteTemplateMappings = list<string * char>
+type MetadataList = obj list
 
 type Endpoint =
-    | SimpleEndpoint   of HttpVerb * RouteTemplate * HttpHandler
-    | TemplateEndpoint of HttpVerb * RouteTemplate * RouteTemplateMappings * (obj -> HttpHandler)
-    | NestedEndpoint   of RouteTemplate * Endpoint list
+    | SimpleEndpoint   of HttpVerb * RouteTemplate * HttpHandler * MetadataList
+    | TemplateEndpoint of HttpVerb * RouteTemplate * RouteTemplateMappings * (obj -> HttpHandler)  * MetadataList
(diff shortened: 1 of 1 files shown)


src/Giraffe/EndpointRouting.fs
```diff
@@ -220,68 +221,76 @@ let routef
...
+    | TemplateEndpoint(v, p, m, h, ml) -> TemplateEndpoint(v, p, m, (fun (o: obj) -> h o >=> httpHandler), ml)
+    | NestedEndpoint(t, lst, ml)       -> NestedEndpoint(t, List.map (applyAfter httpHandler) lst, ml)
+
+let rec addMetadata
+    (metadata: obj)
+    (endpoint: Endpoint) =
+    match endpoint with
+    | SimpleEndpoint(v, p, h, ml)      -> SimpleEndpoint(v, p, h, metadata::ml)
+    | TemplateEndpoint(v, p, m, h, ml) -> TemplateEndpoint(v, p, m, h, metadata::ml)
+    | NestedEndpoint(t, lst, ml)       -> NestedEndpoint(t, lst, metadata::ml)
```
slang25, line 254: This PR all looks great. With this `addMetadata` function, how are we intending it to be used when composing an app:
```fsharp
GET => route "/" (text "Hello World") |> addMetadata something
```
where `something` is an object that describes the whole `HttpHandler` side. Or would there be the ability to use it in the DSL at the front before the handler part? Not that it would be better, I'm just wondering if this PR dictates the usage yet.

Krzysztof-Cieslak: `GET >> addMetadata something => route "/test" (text "Hello World")` should also work (because both `GET` and `addMetdata something` are `Endpoint -> Endpoint`

slang25: Ah, makes sense. I was sleepy and my signature foo was off 😆

Krzysztof-Cieslak:
Ping

PR #440, opened 2020-08-10
Quick fix on documentation

fpellet:
Please to see commits


DOCUMENTATION.md
```diff
@@ -172,7 +172,7 @@ The `task {}` CE is an independent project maintained by [Robert Peele](https://
 If you have `do!` bindings in your Giraffe web application then you must open the `FSharp.Control.Tasks.ContextInsensitive` namespace to resolve any type inference issues:
 
 ```fsharp
-open FSharp.Control.Tasks.V2.ContextInsensitive
+open FSharp.Control.Tasks.ContextInsensitive
```
TheAngryByrd, line 175: TaskBuilder still requires V2. https://github.com/rspeele/TaskBuilder.fs#whats-the-deal-with-the-v2-module

fpellet: I'm not sure I understand. V2 use new implementation, but it's not compatible with old compiler. So, in this case, it's possible to use old implementation, without V2 suffix.
Why it's required to open V2 ? 
And the order is important (use same task variable name). So, I should open V1 after V2 or V2 after V1 ?

TheAngryByrd: Oh you know what I didn't read the whole section that this code was referring to. Now i'm even confused 🤕 .  Definitely need @dustinmoris to chime in here.

fpellet: if that's really what it is, then I don't see what "important notice" brings. Might as well delete the paragraph.

dustinmoris:
Thanks for your PR. I've essentially applied both those mini doc changes to the latest develop branch. Thanks for raising the inconsistencies!

PR #450, opened 2020-11-29
Easier way to composition endpoints

dustinmoris:
Proposal to introduce small API change in the new Endpoint Routing module
---

I have introduced a new union case called `MultiEndpoint` which accepts an `Endpoint list` and re-factored existing functions to accommodate this new union case. I think this new union case makes writing new routes a lot easier.

Before:

```fsharp
let endpoints =
    [
        GET => route "/a" (text "A")
        GET => route "/b" (text "B")
        GET => route "/c" (text "C")
        GET => route "/d" (text "D")
    ]
```

After:

```fsharp
let endpoints =
    [
        GET [
            route "/a" (text "A")
            route "/b" (text "B")
            route "/c" (text "C")
            route "/d" (text "D")
        ]
    ]
```

As a result I also suggest to remove the `=>` operator. I think this operator is just confusing to starters and unnecessary in the context of the new endpoint routing API. It makes it less obvious that the endpoint routes are a flat list and the operator is not applicable the same way as the fish operator in the default router.

So if one would just create a single route they would have to create a single item list.

Before:

```fsharp
GET => route "/foo" (text "bar")
```

After:

```fsharp
GET [ route "/foo" (text "bar") ]
```

Additionally I have also created a new private helper functions `applyHttpVerbsToEndpoints` in order to enable things like `GET_HEAD`:

```fsharp
let endpoints =
    [
        GET [
            route  "/" (text "Hello World")
            routef "/%s/%i" handler2
            routef "/%s/%s/%s/%i" handler3
        ]
        GET_HEAD [
            route "/foo" (text "Bar")
            route "/x"   (text "y")
            route "/abc" (text "def")
        ]
    ]
```

What does everyone think of those changes?


TheAngryByrd (approved):

slang25 (approved):
I think reducing custom operators is a good move. This is an improvement IMO

Krzysztof-Cieslak (approved):

Issue #483, opened 2021-06-06
setContentType function


Choc13:
I've found myself searching for a `setContentType` function, like `setStatusCode` that is just a functional version of the `SetStatusCode` extension method on `HttpContext`. 

I couldn't find this one, am I right in thinking it doesn't exist?

If not would you be happy for me open a PR to add it to `Core.fs` alongside `setStatusCode`?

I believe the implementation is just

```fsharp
let setContentType contentType : HttpHandler =
    fun next ctx ->
        ctx.SetContentType contentType
        next ctx
```

esbenbjerre:
@64J0 I believe this issue can be closed.

PR #484, opened 2021-06-07
Add setContentType function.

Choc13:
As mentioned in #483 it would be nice to have a functional wrapper around `ctx.SetContentType` similar to `setStatusCode`.


dustinmoris (approved):
LGTM

PR #496, opened 2021-11-09
.NET 6

slang25:
A few changes to take advantage of .NET 6 & ASP.NET Core 6 for a Giraffe 6 release.

- Target .NET 6
- Removed Ply (you served us well) and used baked in F# 6 support for tasks
- Updated various dependencies
- Utilise some of the minimal APIs for bootstrapping the included app

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -153,7 +153,7 @@ let webApp =
 Another important aspect of Giraffe is that it natively works with .NET's `Task` and `Task<'T>` objects instead of relying on F#'s historic `async {}` workflows. The main benefit of this is that it removes the necessity of converting back and forth between tasks and async workflows when building a Giraffe web application (because ASP.NET Core only works with tasks out of the box).
 
 ### Tasks - Giraffe 6 and later
-Giraffe 6 targets .NET 6, and utilies F# 6. This means we are able to use [F#'s built-in task support](https://docs.microsoft.com/en-us/dotnet/fsharp/whats-new/fsharp-6#task-), and do not require any additional dependencies.
+Giraffe 6 targets .NET 6 and uses [F# 6's built-in task support](https://docs.microsoft.com/en-us/dotnet/fsharp/whats-new/fsharp-6#task-) without any additional dependencies.
 
 When building web apps using Giraffe, we recommend you use this built-in support too.
 
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -150,13 +150,17 @@ let webApp =
 ```
 
 ### Tasks
+Another important aspect of Giraffe is that it natively works with .NET's `Task` and `Task<'T>` objects instead of relying on F#'s historic `async {}` workflows. The main benefit of this is that it removes the necessity of converting back and forth between tasks and async workflows when building a Giraffe web application (because ASP.NET Core only works with tasks out of the box).
 
-Another important aspect of Giraffe is that it natively works with .NET's `Task` and `Task<'T>` objects instead of relying on F#'s `async {}` workflows. The main benefit of this is that it removes the necessity of converting back and forth between tasks and async workflows when building a Giraffe web application (because ASP.NET Core only works with tasks out of the box).
+### Tasks - Giraffe 6 and later
+Giraffe 6 targets .NET 6, and utilies F# 6. This means we are able to use [F#'s built-in task support](https://docs.microsoft.com/en-us/dotnet/fsharp/whats-new/fsharp-6#task-), and do not require any additional dependencies.
 
-For this purpose Giraffe uses the `task {}` computation expression from the [Ply](https://www.nuget.org/packages/Ply/) NuGet package. Syntactically it works identical to F#'s async workflows (after opening the `FSharp.Control.Tasks` module):
+When building web apps using Giraffe, we recommend you use this built-in support too.
+
+### Tasks - Giraffe 5
+In Giraffe 5, we use the `task {}` computation expression from the [Ply](https://www.nuget.org/packages/Ply/) NuGet package. Syntactically it works identical to F#'s async workflows (after opening the `FSharp.Control.Tasks` module):
 
 ```fsharp
-open FSharp.Control.Tasks
 open Giraffe
 
 let personHandler =
@@ -171,7 +175,7 @@ The `task {}` CE is an independent project maintained by [Crowded](https://githu
 
 **IMPORTANT NOTICE**
 
-If you have `do!` bindings in your Giraffe web application then you must open the `FSharp.Control.Tasks` namespace to resolve any type inference issues:
+If you have `do!` bindings in your Giraffe 5 web application then you must open the `FSharp.Control.Tasks` namespace to resolve any type inference issues:
 
 ```fsharp
 open FSharp.Control.Tasks
@@ -212,8 +216,6 @@ Because an `HttpHandler` is defined as `HttpFunc -> HttpContext -> HttpFuncResul
 The most verbose version of defining a new `HttpHandler` function is by explicitly returning a `Task<HttpContext option>`. This is useful when an async operation needs to be called from within an `HttpHandler` function:
 
 ```fsharp
-open FSharp.Control.Tasks
-
 type Person = { Name : string }
 
 let sayHelloWorld : HttpHandler =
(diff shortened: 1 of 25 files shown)


forki:
Awe!

forki:
Looking good

forki (approved):

NinoFloris (dismissed):

kdblocher:
@slang25 Are there any specific edge cases that should be under test because of swapping out Ply with F#6? Task implementations can be flaky in areas one would not expect.

slang25:
Good question @kdblocher. I wouldn't say flaky, but complex for sure. I consider the Task and the underlying state machine / workflow to be a low level abstraction, which I somewhat understand (not as well as @NinoFloris).

It's an abstraction in that we could use some custom implementation in Giraffe and internalize it (as was done with the very early versions of Giraffe), and consuming apps could use their own implementations of task CEs and it will all "just work".

However it's the implementation of the async workflows, and now F# 6 resumable code and task ce code that really need to be tested, which arguably in F# 6 has been done thoroughly. At least enough to satisfy me.

I'd like to get an updated version of the TechEmpower benchmarks running a beta version of Giraffe with this change, I'd want to see that there is no perf regression and hopefully a nice little boost in numbers.

forki:
As soon as an updated giraffe hits nuget I'll update the techempower

kdblocher:
> I'd like to get an updated version of the TechEmpower benchmarks running a beta version of Giraffe with this change, I'd want to see that there is no perf regression and hopefully a nice little boost in numbers.

This would be a good start.

> Good question @kdblocher. I wouldn't say flaky, but complex for sure. I consider the Task and the underlying state machine / workflow to be a low level abstraction, which I somewhat understand (not as well as @NinoFloris).

Some colleagues and I did a lot of work around this a couple years ago. Our result on the deep dive is more or less "use it the right way, and there's pretty much one or two right ways to do it, and many ways not to. Hopefully F#6 `task` takes care of all of this, too, especially with respect to cancellation and exceptions.

> It's an abstraction in that we could use some custom implementation in Giraffe and internalize it (as was done with the very early versions of Giraffe), and consuming apps could use their own implementations of task CEs and it will all "just work".

I doubt this is necessary, though it would be nice to be able to support interchangeable `task` and `async` scenarios, but that's a lot of work to get the experience right. We don't have HKTs, but SMTCs might work; however, the type checker might spit out unintelligible errors as you sometimes get working with transformers in F#+. (This is obviously out of scope of this PR)

NinoFloris:
@kdblocher I'm the author of Ply and was involved in some of the testing and design of the resumable code feature (as well as the task implementation built on it) and I'm pretty confident in the implementation. 

For the time being the builtin in F# is missing ValueTask return type support so we will still break libraries or people depending on that functionality existing transitively, my opinion is that it's not so important as major versions of Giraffe are only available in one tfm regardless. These projects would all need to add Ply to their direct dependencies instead. Beyond that we may still find some other differences around type inference like https://github.com/dotnet/fsharp/issues/12184 but there haven't been any reports suggesting functional issues.

dustinmoris (approved):
T H A N K S
😃 👏 🥳

TheAngryByrd (approved):

forki:
is there already a prerelease?

dustinmoris:
Yes :)
https://github.com/giraffe-fsharp/Giraffe/actions/runs/1443366693

forki:
Xoxo

Dustin Moris Gorski ***@***.***> schrieb am Mi., 10. Nov.
2021, 10:05:

forki:
https://github.com/TechEmpower/FrameworkBenchmarks/pull/6891 looking good

dsyme:
This is great work. Can anyone link an updated tech-empower result when the figures are available?  Just for clarity?  

Note there was one known bug in the F# 6 `task` implementation, now fixed, here: https://github.com/dotnet/fsharp/issues/12359.    The fix will be in the next .NET 6/VS 2022 release.  The bug is fairly easily avoidable, but still it's probably worth waiting until that fix is through the system before taking this new release of Giraffe out of pre-release.

kdblocher:
@slang25
> However it's the implementation of the async workflows, and now F# 6 resumable code and task ce code that really need to be tested, which arguably in F# 6 has been done thoroughly. At least enough to satisfy me.
> 
> I'd like to get an updated version of the TechEmpower benchmarks running a beta version of Giraffe with this change, I'd want to see that there is no perf regression and hopefully a nice little boost in numbers.

@dsyme 
> Note there was one known bug in the F# 6 `task` implementation, now fixed, here: [dotnet/fsharp#12359](https://github.com/dotnet/fsharp/issues/12359). The fix will be in the next .NET 6/VS 2022 release. The bug is fairly easily avoidable, but still it's probably worth waiting until that fix is through the system before taking this new release of Giraffe out of pre-release.

Ditto that this upgrade happening so quickly is great, but it's exactly this sort of thing ([dotnet/fsharp#12359](https://github.com/dotnet/fsharp/issues/12359)) that is the stuff of nightmares for package maintainers.

Do the TechEmpower benchmarks happen to additional provide good runtime coverage so that we could discover more edge-case bugs like this, or does another suite of tests somewhere exist for that?

dsyme:
@kdblocher Yes, understood.

> Do the TechEmpower benchmarks happen to additional provide good runtime coverage so that we could discover more edge-case bugs like this, or does another suite of tests somewhere exist for that?

dotnet/fsharp has an extensive test suite of course, though the more real-world code the better.

slang25:
I certainly tempted fate with my remarks 😆 

The techempower benchmarks have a few simple and well defined use cases, running under great load. I wouldn't expect it to surface this sort of bug (although it could), it's more likely to surface performance regressions.

Our current approach is still appropriate, we have alpha packages out that we can use to trial and gauge our confidence in the changes. We've got a clear signal that there's a messy corner case bug, so we'll hold off. Thanks @dsyme for the heads-up.

slang25:
@dsyme I think we're waiting on this set of results:
https://tfb-status.techempower.com/results/939169a8-6a56-44ae-9e50-0e714753c445

I'm also looking to enable GPO just like the C# ASP.NET Core versions: https://github.com/TechEmpower/FrameworkBenchmarks/pull/6923

dsyme:
> Our current approach is still appropriate, we have alpha packages out that we can use to trial and gauge our confidence in the changes. We've got a clear signal that there's a messy corner case bug, so we'll hold off. Thanks @dsyme for the heads-up.

Yes, I agree, the current approach is appropriate.  Note users always have the opportunity to use `Ply` or `TaskBuilder` for their own tasks should another corner-case crop-up.

> @dsyme I think we're waiting on this set of results:

Thanks. Please let me know if you see any specific significant regressions, we can likely turn-around a fix fairly quickly.  Ply has really impressive performance so I'm expecdting some +/- differences.

dsyme:
@slang25 Hmmmm I'm concerned there might be a perf slow down here:

Taking the top two here: https://tfb-status.techempower.com/

Before:

image

After:

image

We can wait for the full results but that looks problematic.  If that persists then we'll need a standalone test that we can profile and understand

DOCUMENTATION.md
```diff
@@ -171,7 +175,7 @@ The `task {}` CE is an independent project maintained by [Crowded](https://githu
 
 **IMPORTANT NOTICE**
 
-If you have `do!` bindings in your Giraffe web application then you must open the `FSharp.Control.Tasks` namespace to resolve any type inference issues:
+If you have `do!` bindings in your Giraffe 5 web application then you must open the `FSharp.Control.Tasks` namespace to resolve any type inference issues:
```
dsyme, line 178: This should be removed I think since it's only relevant to Giraffe 5

slang25:
Yeah, I saw that and hoped it was something to do with it being a partial result, a bit concerned here

dustinmoris:
Let's see what happens but I think I once observed something similar when all results were not in yet.

dsyme:
Either way it's probably better to phase this more carefully. First move to net6, then remove ply. Probably best to get that on the works actually. 

I'm sure we'll get to the bottom of it, even if a fix is required in net sdk 6.0.200/dev 17.1. And to be honest it's really amazing to have such end to end benchmarking available. 

Note we've done plenty of raw micro benchmarking and things are OK, so if it's a glitch it will not be fundamental, certainly not that much off.

NinoFloris:
It might just be logging 😅

https://github.com/TechEmpower/FrameworkBenchmarks/pull/6927/files

dsyme:
Yay! Thanks for finding that!

forki:
.NET SDK 6.0.200 is out. Everything looks good now. I think new Giraffe can be released!

PR #499, opened 2021-11-16
Fix endpoint routing docs

semyon2105:
Operator `=>` got removed but it's still being used in a code sample in the docs


dustinmoris (approved):
Thank you!

PR #502, opened 2021-12-31
feat: add request limits on accept, content-type, and content-length headers

### LINKED ISSUE #392: Feature request - request limitation
**Is your feature request related to a problem? Please describe.**
Provide basic request limitations, in the context of a more secure/safe API implementations out of the box.

**Describe the solution you'd like**
Some extra middleware functions that limit requests based on header and body information, using the existing HTTP feature for the body size.

```fsharp
/// Functionality to limit HTTP requests based on header and body information.
[<AutoOpen>]
module Limit =
  /// Limits to only requests with one of the specified `Content-Type` headers,
  /// returning `406 NotAcceptable` when the request header doesn't exists in the set of specified types.
  let contentTypes contentTypes =
    fun next (ctx : HttpContext) ->
      match Option.ofObj ctx.Request.ContentType with
      | Some t when Seq.contains t contentTypes -> next ctx
      | Some _ -> RequestErrors.notAcceptable (text "request header 'Content-Type' doesn't had expected value") earlyReturn ctx
      | None -> RequestErrors.notAcceptable (text "expected to have request header 'Content-Type'") earlyReturn ctx

  /// Limits to only requests with a specific `Content-Type` header, 
  /// returning `406 NotAcceptable` when the request header value doesn't match the specified type.
  let contentType contentType = contentTypes [ contentType ]

  /// Limits request `Content-Length` header to a specified length, 
  /// returning `406 NotAcceptable` when no such header is present or the value exceeds the maximum specified length.
  let contentLength l =
    fun next (ctx : HttpContext) ->
      match Option.ofNullable (ctx.Request.GetTypedHeaders().ContentLength) with
      | Some v when v <= l ->  next ctx
      | Some _ -> RequestErrors.notAcceptable (text "request header 'Content-Length' is too large") earlyReturn ctx
      | None -> RequestErrors.notAcceptable (text "request doesn't contain 'Content-Length' header") earlyReturn ctx

  /// Limits the request body size to a specified length, returing `413 Payload Too Large` if the body size exceeds the specified maximum.
  let maxBodySize l =
    fun next (ctx : HttpContext) ->
      ctx.Features.Get<IHttpMaxRequestBodySizeFeature>().MaxRequestBodySize <- Nullable<_> l
      try next ctx
      with :? BadHttpRequestException as ex when ex.StatusCode = 413 ->
        (clearResponse >=> setStatusCode ex.StatusCode >=> text "request body too large") earlyReturn ctx
```

Is this too 'application-specific' to be added to Giraffe? I guess since we have a `mustAccept` core function, we could have some extra's as well. All in the context of having some boilerplate code to make the application more safer by default.

stijnmoreels:
Adds request guards/limitation on:
* `Accept` header
* `Content-Type` header
* `Content-Length` header
With user-friendly error http responses.

Closes #392


stijnmoreels:
I tried to follow the style guidelines and way of working, please let me know if I did anything wrong. Happy to help! 👍

TheAngryByrd (commented):
Looks good from my end.  I only have some questions that other maintainers (@dustinmoris) should weigh in on.

So far Giraffe doesn't have many error messages it returns that are hard coded.  I could [only find one place](https://github.com/giraffe-fsharp/Giraffe/blob/f4792838a57ecfb1de6987745fc66b8654ac5f53/src/Giraffe/Negotiation.fs#L39) where we do it.  I'm wary of having error messages like this hard coded for localization and even preference reasons. However I'm unsure of how to resolve this easily within this library.

stijnmoreels:
> Looks good from my end. I only have some questions that other maintainers (@dustinmoris) should weigh in on.
> 
> So far Giraffe doesn't have many error messages it returns that are hard coded. I could [only find one place](https://github.com/giraffe-fsharp/Giraffe/blob/f4792838a57ecfb1de6987745fc66b8654ac5f53/src/Giraffe/Negotiation.fs#L39) where we do it. I'm wary of having error messages like this hard coded for localization and even preference reasons. However I'm unsure of how to resolve this easily within this library.

Maybe we can have a different set of functions that takes in a custom error HTTP handler. There maybe needed two, as we check both on the presence of the header and whether the header has the correct value.

Banashek:
A few different solutions exist

### 1) Defer Localization to later

Pros: 

* Less work now
* Can create a separate issue, note down current hardcoded values (including the ones being introduced in this PR) and handle separately

Cons: 

* More work later
* Currently hardcoded values will need to adapt
* Risk of deprioritization and continued non-localization in code

### 2) Leverage Aspnet Core builtins

https://docs.microsoft.com/en-us/aspnet/core/fundamentals/localization?view=aspnetcore-6.0

A [RequestLocalization Middleware](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.builder.applicationbuilderextensions.userequestlocalization?view=aspnetcore-6.0) could be registered, with a default `.resx` file (or binary `.resources` file)  in a `i18n` directory somewhere perhaps?

Pros:

* The middleware seems fairly well designed and I imagine that it's been tested in production applications to influence its design
* Follows a fairly established pattern for i18n, and since Giraffe already follows aspnetcore-isms, registering a middleware with a default location would not be much work

Cons:

* May require designing an escape hatch for if someone doesn't want to include the middelware (custom i18n solution, etc)
* Code could not be the most functional/ergonomic, may require some wrapping for functional use 

### 2) Custom functional i18n 

Something like [Oldschooli18n](https://github.com/d-p-y/oldschool-i18n) or an alternate functional i18n approach

Pros:

* Would potentially allow for a more ergonomic and testable approach

Cons:

* More code to write/test

---

I'm somewhat in favor of 1, then one of 2/3, so that this PR can get merged, but that the goal of i18n isn't lost.

dustinmoris:
I'd prefer to make the first parameter of each of these handlers to accept an error handler. It's not just about changing the message, but the entire response should be flexible. For example an API could return this as a JSON object or plain text response but a website might want to render a HTML page with the error displayed.

Here's an example of where this approach has already been taken:
https://github.com/giraffe-fsharp/Giraffe/blob/master/src/Giraffe/Negotiation.fs#L159

Of course in the final user application one would define that handler only once and then use the new handler for everything else, e.g.:

```
let someErrorHandler = ...

let validateContentType = haveAnyContentTypes someErrorHandler contentTypes

// then use validateContentType everywhere else
```

stijnmoreels:
Ok, that's fine by me! 👍

64J0:
This feature looks pretty cool!

64J0:
Just a first heads-up, please target the `develop` branch instead of `master`. The workflow is documented here: https://github.com/giraffe-fsharp/Giraffe?tab=readme-ov-file#contributing.

DOCUMENTATION.md
```diff
@@ -2377,6 +2378,50 @@ let webApp =
...
+**Content-Type**
+Guards http request based on its `Content-Type` header:
+
+```fsharp
+// Only allow http request with a `Content-Type` header `equals `application/json`.
+let webApp =
+  GET >=> haveContentType "application/json" >=> text "Hello World"
+
+// Http request with    `Content-Type` = `application/json`   -> pass through
+// Http request without `Content-TYpe` = `application/json`   -> error 406 http response: "cannot accept request because expected to have request header 'Content-Type'"
```
64J0, line 2408:
```suggestion
// Http request without `Content-Type` = `application/json`   -> error 406 http response: "cannot accept request because expected to have request header 'Content-Type'"
```

64J0:
I'll review this PR either this or next week.

64J0:
I changed this branch's code to accept custom handlers from the user of this feature. And added more unit tests to validate these new scenarios.

nojaf (approved):
Sorry for the late review here. This slipped off my radar and I don't feel very qualified to say much about this. I don't use Giraffe that often and that makes it hard to have any opinion about the API.

I would merge this in and publish an alpha.
Ask people if the API works for them and take it from there.

Issue #485, opened 2021-06-18
Accept: */* header does not pass mustAccept


avallee:
Using the handler mustAccept ["text/html"] with a request with the exact accept header: "*/*" fails by skipping the pipeline.

According to rfc7231 Section 5.3.2

> The asterisk "*" character is used to group media types into ranges,
> with "*/*" indicating all media types and "type/*" indicating all
> subtypes of that type.  The media-range can include media type
> parameters that are applicable to that range.

Banashek:
This does look to be broken.
Fortunately, we can do the Giraffe thing and lean on the [MediaTypeHeaderValue](https://github.com/dotnet/aspnetcore/blob/0ee742c53f2669fd7233df6da89db5e8ab944585/src/Http/Headers/src/MediaTypeHeaderValue.cs) class which has a [isSubsetOf](https://github.com/dotnet/aspnetcore/blob/0ee742c53f2669fd7233df6da89db5e8ab944585/src/Http/Headers/src/MediaTypeHeaderValue.cs#L385) method that should help in this scenario.

I've created a PR with tests I'll add.

PR #516, opened 2022-07-06
Updating mimetype accept header parsing to use builtin aspnet parse/methods

Banashek:
Fixes [485](https://github.com/giraffe-fsharp/Giraffe/issues/485)

This change updates the `mustAccept` to base it's decisions on the values of parsed [MediaTypeHeaderValue](https://github.com/dotnet/aspnetcore/blob/0ee742c53f2669fd7233df6da89db5e8ab944585/src/Http/Headers/src/MediaTypeHeaderValue.cs) values instead of strings.

It utilizes the [isSubsetOf](https://github.com/dotnet/aspnetcore/blob/0ee742c53f2669fd7233df6da89db5e8ab944585/src/Http/Headers/src/MediaTypeHeaderValue.cs#L385) method of that class in order to do proper subset matching.

While the parsing is wrapped in an exception handler during the runtime execution, the construction of the function does not wrap the parsing. This is so that it will fail fast on startup (developer supplied values), but tolerate malformed values at runtime (bad user input).

I've added only a couple of tests, but believe it should suffice for this scenario since the code leans on the Net builtins.

Happy to make changes (remove rfc comments, etc) if desired.

### DIFF
--- a/src/Giraffe/Core.fs
+++ b/src/Giraffe/Core.fs
@@ -7,6 +7,7 @@ module Core =
     open System.Globalization
     open Microsoft.AspNetCore.Http
     open Microsoft.Extensions.Logging
+    open Microsoft.Net.Http.Headers
     open Giraffe.ViewEngine
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


dustinmoris (approved):
LGTM, many thanks!

Issue #513, opened 2022-05-15
Integer Overflow in the versioning


aloisdg:
Hello,

The current versioning logic is going to run into an integer overflow. The last pack is `2139995418` so we are still not at `2147482963` but we will run into. Here is how we handle it in [Cardizer](https://github.com/d-edge/Cardizer/pull/124):

We went from

    version=$GITHUB_RUN_ID
            
to 

```bash            
git fetch --prune --unshallow --tags --quiet
latestTag=$(git describe --tags --abbrev=0 2>/dev/null || echo 0.0.1)
runId=$GITHUB_RUN_ID
version="${latestTag//v}-build.${runId}"
echo Non-release version: $version
```
    
The goal is to grab the latest tag in the current branch and append it the RUN_ID as metadata.

I would gladly hear any comment about this solution.

Have a nice day.

Banashek:
The solution above sounds like it would work fine.

My understanding is that this version specified in the build file [here](https://github.com/giraffe-fsharp/Giraffe/blob/6fe39c21abdf85bf1e49db55a053342c42186e7f/.github/workflows/build.yml#L43) really is just for a unique running id related to the build for packaging non-release versions.

As per the mention from $514 it does appear to be causing problems already:
https://github.com/giraffe-fsharp/Giraffe/runs/7060792905?check_suite_focus=true#step:7:31

`/home/runner/.dotnet/sdk/6.0.100/Sdks/NuGet.Build.Tasks.Pack/build/NuGet.Build.Tasks.Pack.targets(110,5): error MSB4018: System.ArgumentException: PackageVersion string specified '2564178360' is invalid. [/home/runner/work/Giraffe/Giraffe/src/Giraffe/Giraffe.fsproj]`

The proposed solution would change the version from `2564178360` to `6.0.0-build.2564178360`, which appears like it would work as long as the git tags continue to follow the `vX.Y.Z` format.

aloisdg:
@Banashek The solution was adopted by our team after some [discussions](https://github.com/d-edge/Cardizer/pull/124).

Banashek:
Ah yeah a lot more thoughtful discussion in that thread!
Thanks for bringing that knowledge over.
Hope you don't mind the quick PR, just wanted to remove as many barriers as possible for maintainers.

TheAngryByrd:
https://github.com/giraffe-fsharp/Giraffe/pull/517

PR #517, opened 2022-07-06
Updating github workflow file to use nuget-acceptable version numbers for packaging

Banashek:
Fixes [513](https://github.com/giraffe-fsharp/Giraffe/issues/513) and should unblock builds.

This change alters the versioning used by github actions for builds by utilizing the Github Run Identifier as an "Additional Label" (in semver parlance) rather than a major version number.

Example (from [this failed build](https://github.com/giraffe-fsharp/Giraffe/runs/7060792905?check_suite_focus=true))

* Before: `2564178360` _(`System.ArgumentException: PackageVersion string specified '2564178360' is invalid.`)_
* After: `6.0.0-build.2564178360`

This matches the semver logic: `Example: 1.0.0-alpha < 1.0.0-alpha.1 < 1.0.0-alpha.beta < 1.0.0-beta < 1.0.0-beta.2 < 1.0.0-beta.11 < 1.0.0-rc.1 < 1.0.0.` (ref: https://semver.org/)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -40,7 +40,11 @@ jobs:
         run: dotnet test -c Release
       - name: Pack
         if: matrix.os == 'ubuntu-latest'
-        run: dotnet pack -v normal -c Release --no-restore --include-symbols --include-source -p:PackageVersion=$GITHUB_RUN_ID src/$PROJECT_NAME/$PROJECT_NAME.*proj
+        run: |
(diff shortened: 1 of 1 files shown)


dustinmoris (approved):
Thanks for that! I am extremely annoyed that .NET keeps changing shit like this all the time. This was fine in previous .NET Core versions and now they don't accept non-SemVer compatible versions anymore which is why this war broken. Appreciate the fix!

PR #522, opened 2022-08-28
[WIP] Remove Kleisli composition in favour of plain composition and pipelining

dbrattli:
This PR is a work in progress experiment to use plain composition and pipelining instead of Kleisli composition for composing Giraffe HTTP handlers. The goal is to be transparent about this effort, and to get comments and suggestions on how I could improve it further.

Using Kleisli with F# can be problematic since Kleisli can be a difficult FP abstraction to understand and also collides easily with other libraries using Kleisli. Operators such as `>>` and `|>` works for all data types, and also have better IDE support.

This PR enables the HTTP handlers to compose themselves by adding a `source: HttpHandler` as the last argument to every  "operator". This is similar to how abstractions like e.g `Observable` works. This means that the main abstraction is the `HttpHandler` and you have "operators" (`HttpHandler -> HttpHandler`) that transforms the `HttpHandler`. So the only difference from the original Giraffe is that there is no explicit `compose` i.e `>=>` but that the operators compose themselves when applied.

Example:

Code that currently needs to be written using `>=>` such as:

```fs
let someHttpHandler : HttpHandler =
    setStatusCode 200
    >=> text "Hello World"
```

Would now in this PR be written as:

```fs
let someHttpHandler : HttpHandler -> HttpHandler =
    setStatusCode 200
    >> text "Hello World"
```

I have also added upper case versions of handlers so they can be used as initial handlers and thus enable pipelining:

```fs
let someHttpHandler : HttpHandler =
    SET_STATUS_CODE 200
    |> text "Hello World"
```

Naming is hard so unsure if I should use uppercase names (aligns a bit with the current design) or use ticked names e.g `setStatusCode'` for initial handlers. In other libraries you usually have unique handlers for the start of the pipeline but with Giraffe you can use most handlers both initially and in the middle of a pipeline. Other ideas how this could be done in a better way?

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -107,9 +107,9 @@ There is no limit to how many `HttpHandler` functions can be chained with `compo
 ```fsharp
 let app =
     route "/"
-    >=> setHttpHeader "X-Foo" "Bar"
-    >=> setStatusCode 200
-    >=> setBodyFromString "Hello World"
+    >> setHttpHeader "X-Foo" "Bar"
+    >> setStatusCode 200
+    >> setBodyFromString "Hello World"
 ```
 
 If you would like to learn more about the `>=>` (fish) operator then please check out [Scott Wlaschin's blog post on Railway oriented programming](http://fsharpforfunandprofit.com/posts/recipe-part2/).
@@ -121,8 +121,8 @@ The `choose` combinator function iterates through a list of `HttpHandler` functi
 ```fsharp
 let app =
     choose [
-        route "/foo" >=> text "Foo"
-        route "/bar" >=> text "Bar"
+        route "/foo" >> text "Foo"
+        route "/bar" >> text "Bar"
     ]
 ```
 
@@ -144,8 +144,8 @@ let time() = System.DateTime.Now.ToString()
 
 let webApp =
     choose [
-        route "/normal"  >=> text (time())
-        route "/warbler" >=> warbler (fun _ -> text (time()))
+        route "/normal"  >> text (time())
+        route "/warbler" >> warbler (fun _ -> text (time()))
     ]
 ```
 
@@ -380,8 +380,8 @@ open Giraffe
 
 let webApp =
     choose [
-        route "/ping"   >=> text "pong"
-        route "/"       >=> htmlFile "/pages/index.html" ]
+        route "/ping"   >> text "pong"
+        route "/"       >> htmlFile "/pages/index.html" ]
 
 type Startup() =
     member __.ConfigureServices (services : IServiceCollection) =
@@ -414,8 +414,8 @@ open Giraffe
 
 let webApp =
     choose [
-        route "/ping"   >=> text "pong"
-        route "/"       >=> htmlFile "/pages/index.html" ]
+        route "/ping"   >> text "pong"
+        route "/"       >> htmlFile "/pages/index.html" ]
 
 let configureApp (app : IApplicationBuilder) =
     // Add Giraffe to the ASP.NET Core pipeline
@@ -730,8 +730,8 @@ let notFoundHandler : HttpHandler =
 
 let webApp =
     choose [
-        route "/foo" >=> text "Foo"
-        route "/bar" >=> text "Bar"
+        route "/foo" >> text "Foo"
+        route "/bar" >> text "Bar"
         notFoundHandler
     ]
 ```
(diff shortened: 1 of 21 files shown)


isaacabraham:
What are the effects in terms of usage and common patterns in Giraffe - are we talking about simply removing the Kleisi, replacing with pipeline and everything else carries on as normal - how would this impact the average developer using Giraffe and / or Saturn?

dsyme:
Can you make it activated by an 'open' and the names stay as they are?

7sharp9:
`open sesame`  I double dare you.

LiteracyFanatic:
Here are my thoughts as someone who uses Giraffe pretty heavily.

I appreciate that the F# community has largely rejected the abuse of custom operators for everything under the sun that some other functional languages fall into. Because library specific operators have no inherent meaning tied to them, it becomes difficult to even guess what code does without either already being familiar with the library or consulting its documentation.

That being said, Kleisli composition is one of the easier operators to understand in my opinion and has a pretty firmly established history in the F# web programming space. There are a lot of existing tutorials and code samples written in the current style. Plus it's the only operator being used by Giraffe, so it's not like people will have difficulty picking the correct operator for their use case (a very real concern with something like FParsec).

Another concern I have is that when writing a custom `HttpHandler` it is possible to forget to pipe to `source`. This still type checks correctly and depending on the exact type annotations you use may not even trigger an unused value warning. I could see this leading to some subtle and hard to debug errors when part of the handler pipeline just doesn't run.

Also, I find the special starting handlers needed for pipelineing to be really awkward. The existing usage of capitalized handlers for automatic content negotiation is already a bit opaque for my liking, and this would make things even more confusing. The main reason to prefer pipelineing in my mind is that it encourages people to explicitly annotate the types their functions take. When you define an arbitrary function as a composition of other functions, you have to rely on the IDE or your memory to know what arguments it expects and what it will return. This isn't really an issue in this case though because `HttpHandler`s have a fixed shape. If people really want to use the `|>` operator, I think it is best to just expose a single empty handler that would be used at the start instead of adding a bunch of extra functions for every handler you might want to start with.

To clarify, I'm not opposed to using `>>` instead of `>=>` if we truly think it will make it easier for people getting started, just pointing out potential issues I see.

Savelenko:
I agree with @LiteracyFanatic on most points. The motivation for this PR is not really clear, while I suspect that _implementing_ new custom HTTP handlers might be impacted negatively due to the change in the underlying types. I believe @LiteracyFanatic also refers to that with a more concrete example. So while removal of >=> might provide minor perceived convenience to some users, i.e. in straightforward usage by combining predefined handlers, medior/advanced users (i.e. the main target audience) of Giraffe could be impacted negatively in a much more substantial way.

I only looked lightly at the changes in the core types, so can't support this impression more thoroughly. It would be nice to see an analysis of this as part of the PR. I might be wrong.

dbrattli:
I btw think I have found a better way to do this. By adding pipeline enabled "extension" functions that do not interfere with the existing Giraffe, i.e:

```fs
module Pipeline =
    let inline ptext (str : string) (source: HttpHandler) : HttpHandler =
        source >=> text str

    let inline pchoose handlers (source: HttpHandler) : HttpHandler =
        source >=> choose handlers

    let inline pjson<'T> (dataObj : 'T) (source: HttpHandler) : HttpHandler =
        source >=> json dataObj

   ...
```

This let's us write pipeline enabled code by piping int the pipeline-enabled functions:

```fs
    let app =
        GET |> pchoose [
            route "/"     |> ptext "Hello World"
            route "/foo"  |> ptext "bar"
            route "/json" |> pjson { Foo = "john"; Bar = "doe"; Age = 30 }
            setStatusCode 404 |> ptext "Not found" ]
```

This could be added as an optional module to Giraffe itself, or as an extension NuGet e.g `Giraffe.Pipeline`. I will make a new PR within a few days. What do you think?

codingedgar:
`Giraffe.Pipeline` sounds like an excellent idea.

I do share the same concerns as the others, that change is going to break a lot of code, introduce silent errors, and has no justification for mid/advanced users.

But a separate syntax-sugar library is in the best spirit.

If enough adoption it can demonstrate that it is in fact preferred with concrete data.

It is also a common pattern to use simplified adapters from a 3rd party library so novices will be familiarised with the approach.

dbrattli:
Thanks for all the comments, I will be closing this PR now since I got the feedback I needed to make progress. The current thinking is to create a separate companion library for Giraffe called [Giraffe.Pipelines](https://github.com/dbrattli/Giraffe.Pipelines) that enables the use of pipeline operator together with Giraffe. If this works well we can try to get the repo as part of https://github.com/giraffe-fsharp/. Time will tell.

PR #541, opened 2023-07-05
RFC-compliant Content-Length handling for 1xx, 204 and 205 responses and CONNECT requests

### LINKED ISSUE #540: RFC-compliant Content-Length handling for 1xx, 204 and 205 responses and CONNECT requests
Giraffe should make sure to not send non-RFC-compliant Content-Length headers.

Related Kestrel issue:
https://github.com/dotnet/aspnetcore/issues/42956

Related Kestrel PR:
https://github.com/dotnet/aspnetcore/pull/43103

Context:
Since .NET 7, Kestrel throws an exception if Giraffe tries to send a wrong Content-Length header. See PR above.
My server receives quite a few CONNECT requests that now always return 500.

We should probably fix it here in Line 345:
https://github.com/giraffe-fsharp/Giraffe/blob/e206cd74f9adcc1d78bf96f922be9029b986e626/src/Giraffe/HttpContextExtensions.fs#L342-L349

retendo:
Fixes #540


TheAngryByrd (approved):
Thanks!

PR #546, opened 2023-08-07
Add JSON Write Async extension with custom setting

precisemath:
## Description

Create an extension method to add custom JSON serialization settings. There is no way currently to override the default JSON serializer settings.


dustinmoris:
There is a way to change the default settings. Have a look [here](https://giraffe.wiki/docs#json).

Issue #560, opened 2023-10-08
Migrate to System.Text.Json


esbenbjerre:
I think given the updates to System.Text.Json it may be mature enough to [replace Newtonsoft](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/migrate-from-newtonsoft?pivots=dotnet-7-0) in Giraffe.

64J0:
+1

Issue #481, opened 2021-05-31
Remove System.Text.Json NuGet reference


cmeeren:
AFAIK System.Text.Json is always available in .NET 5. Therefore, since Giraffe only targets .NET 5, there is no need for the Giraffe NuGet package to depend on it.

dbrattli:
Fixed by #563

Issue #451, opened 2020-11-29
Move Utf8Json and Newtonsoft into separate packages?


dustinmoris:
Should Giraffe make the new `System.Text` JSON serializer the default and then add support for `Utf8Json` and `Newtonsoft` via two separate NuGet packages?

Most Giraffe applications will only require a single JSON serializer and therefore it feels unnecessary to create a dependency to all three serializers if only one will ever be used.

Trying to collect some feedback before making any decisions.

@TheAngryByrd @slang25 @Krzysztof-Cieslak @JonCanning @jchannon @isaacabraham

Krzysztof-Cieslak:
Sounds like a good idea to me.

isaacabraham:
The UTF8 one (IIRC) has a load of downstream dependencies on stuff like System.ValueTuple which shouldn't be needed on netcore since (I think) netcore2 so pulling that one out at a minimum would be a good thing.

I think most people in the F# world currently will do one of the following:

1. Use whatever the default serializer is. I imagine this will account for the vast majority of Giraffe users, or at least until they hit an issue in terms of serialization.
1. Use a specific serializer - probably Newtonsoft.Json or Thoth. Both work well with F# types.

System.Text.Json didn't seem quite ready for F# usage when [we last looked at it](https://www.compositional-it.com/news-blog/json-serialisation-in-f/), particularly for deserialization. My gut feel says to stick with Newtonsoft.JSON as the default serializer but to remove the UTF8 one (although ironically if I remember looking at the Techempower benchmarks, isn't that the fastest one!).

jchannon:
The one thing I did with Carter was move it to System.Text.Json then had to add back a Json.Net serializer for people to use if they wanted to fall back on it.  

However, I think for F# there are some converters etc that can make STJ workable. From a dependency perspective, STJ would be nicer but obviously there pros/cons in this area

slang25:
I think `System.Text.Json` will never "be ready" for F# on its own, that's what `FSharp.SystemTextJson` provides. I think it's still a good default though, even with the `FSharp.SystemTextJson` caveat, as we get  `System.Text.Json` "for free" as a dependency, in .NET it's part of the `FrameworkReference`.

So maybe we default it to `System.Text.Json`, but the docs make it clear how to extend it with `FSharp.SystemTextJson`, or switch it to `Newtonsoft.Json` or `Utf8Json`?

isaacabraham:
[...]

PR #563, opened 2023-10-14
Promote System.Text.Json as default JSON serializer

esbenbjerre:
## Description

This PR removes the Newtonsoft JSON serializer and promotes System.Text.Json as the default. I think System.Text.Json is the better choice these days. Among other things the default serializer
- Uses camel-case
- Serializes `Option` as `value` or `null`
- Supports ignoring `Option` or `null` values using the `[<JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingDefault)>]` attribute

Any other specific behaviour should IMO be implemented by users, using attributes (e.g., `[<JsonPropertyName("someName")>]`, `[<JsonConverter(typeof<SomeCustomJsonConverter>)>]`), as there's no sensible defaults for most user-defined types.

## How to test

Run tests and try to serialize/deserialize a JSON payload in a `HttpHandler`.

## Related issues

https://github.com/giraffe-fsharp/Giraffe/issues/481
https://github.com/giraffe-fsharp/Giraffe/issues/560
https://github.com/giraffe-fsharp/Giraffe/issues/451


esbenbjerre:
@dustinmoris

TheAngryByrd:
Why are you removing Newtonsoft entirely? I can see maybe swapping the default but not outright removal.

esbenbjerre:
It was my impression from discussion in #451 that it would be preferable to have it in a separate NuGet package and I agree with that sentiment. No reason to include the dependency if it's not being used.

esbenbjerre:
I also think this switch would get us closer to finally having a `tryBindJson` function.

RJSonnenberg (approved):

DOCUMENTATION.md
```diff
@@ -2933,72 +2933,7 @@ Please visit the [Giraffe.ViewEngine](https://github.com/giraffe-fsharp/Giraffe.
...
-let main _ =
-    WebHost.CreateDefaultBuilder()
-        .Configure(Action<IApplicationBuilder> configureApp)
-        .ConfigureServices(configureServices)
-        .ConfigureLogging(configureLogging)
-        .Build()
-        .Run()
-    0
-```
+By default Giraffe uses `System.Text.Json` for (de-)serializing JSON content. An application can modify the default serializer by registering a new dependency which implements the `Json.ISerializer` interface during application startup.
```
64J0, line 2936: TODO: add documentation explaining how someone could configure `Newtonsoft.Json` in case it's still necessary.

src/Giraffe/HttpContextExtensions.fs
```diff
@@ -222,7 +222,7 @@ type HttpContextExtensions() =
     /// Uses the <see cref="Json.ISerializer"/> to deserializes the entire body of the <see cref="Microsoft.AspNetCore.Http.HttpRequest"/> asynchronously into an object of type 'T.
     /// </summary>
     /// <typeparam name="'T"></typeparam>
-    /// <returns>Retruns a <see cref="System.Threading.Tasks.Task{T}"/></returns>
+    /// <returns>Returns a <see cref="System.Threading.Tasks.Task{T}"/></returns>
```
64J0, line 225: Good catch!

DOCUMENTATION.md
esbenbjerre: @64J0 I've added it now.

64J0: Do you mind rebasing this PR too?

DOCUMENTATION.md
```diff
@@ -2933,88 +2933,84 @@ Please visit the [Giraffe.ViewEngine](https://github.com/giraffe-fsharp/Giraffe.
...
 
-    // Now register Utf8Json.Serializer
-    this.AddSingleton<Json.ISerializer>(Utf8Json.Serializer(Utf8Json.Serializer.DefaultResolver)) |> ignore
+        member __.Deserialize<'T> (json : string) = // ...
+        member __.Deserialize<'T> (bytes : byte[]) = // ...
+        member __.DeserializeAsync<'T> (stream : Stream) = // ...
 ```
 
-Or to use `SystemTextJson.Serializer` instead of `NewtonsoftJson.Serializer`, register a new dependency of type `Json.ISerializer` during application configuration:
+For example, one could define a `Newtonsoft.Json` serializer:
```
64J0, line 2955: Nice, thanks for adding this section!

DOCUMENTATION.md
esbenbjerre: Sure thing

DOCUMENTATION.md
```diff
@@ -2933,59 +2933,7 @@ Please visit the [Giraffe.ViewEngine](https://github.com/giraffe-fsharp/Giraffe.
...
-    services.AddGiraffe() |> ignore
-
-    let serializationOptions = SystemTextJson.Serializer.DefaultOptions
-    // Optionally use `FSharp.SystemTextJson` (requires `FSharp.SystemTextJson` package reference)
-    serializationOptions.Converters.Add(JsonFSharpConverter(JsonUnionEncoding.FSharpLuLike))
-    // Now register SystemTextJson.Serializer
-    services.AddSingleton<Json.ISerializer>(SystemTextJson.Serializer(serializationOptions)) |> ignore
-```
-
-#### Customizing JsonSerializerSettings
```
dbrattli, line 2961: Should we really remove this section? How do we customize the JsonSerializerSettings now?

esbenbjerre: Valid point, I've added it back but now below the "Using a different JSON serializer" section.

dbrattli (approved):
Great work. Thanks for doing this!!

64J0:
I'll re-review this PR during this week. Thanks for the updates @esbenbjerre!

TheAngryByrd (approved):
Thanks for the work on this! 

---

@64J0 I want to mention I think this should be considered as a major version bump. Swapping JSON serialization can be a drastic change and shouldn't be taken lightly. A major bump should be a big signal that something has changed. It should also be very loud and clear in release notes with migration links to STJ or how to put back the Newtonsoft serializer.

64J0 (approved):

64J0:
Thanks for the heads-up @TheAngryByrd! Since this PR is already open for a while, and we have many approvals, I'm going to merge it right now. Thanks for the contribution @esbenbjerre, great job!

aspnetde:
JFYI – That turns out to be a massive breaking change to one of my applications which runs on Giraffe since 4 years. Not blaming anyone, but having some change log which makes things like that more explicit than the release page on GH would be nice ✌️

64J0:
Hi @aspnetde, sorry for this breaking change. @abelbraaksma gave some suggestions at [this issue](https://github.com/giraffe-fsharp/Giraffe/issues/606) on how to improve for future updates.

> but having some change log which makes things like that more explicit than the release page on GH would be nice

Can you please expand this comment? What would you like to have found regarding this change?

PR #567, opened 2023-11-28
Minor code optimisation

Thorium:
This is a tiny PR to make the route-check a tiny bit faster by not doing dictionary lookup twice.

### DIFF
--- a/src/Giraffe/Routing.fs
+++ b/src/Giraffe/Routing.fs
@@ -8,9 +8,9 @@ module SubRouting =
     let private RouteKey = "giraffe_route"
 
     let getSavedPartialPath (ctx : HttpContext) =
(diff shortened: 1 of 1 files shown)


Thorium:
Another modification to consider, would be adding `[<Struct>]` attribute to Giraffe's minor enum-types like HttpVerb, MatchMode, Precondition, CacheDirective and RangeBoundary. But if there are proper performance tests somewhere, could be first tested if they'd have any effect or not.

64J0 (approved):
Good catch!

64J0 (changes requested):
Can you please target the `develop` branch instead of the `master`directly? My bad, just noticed it now.

PR #568, opened 2023-11-28
No need to build so many Regex

Thorium:
Another tiny performance optimization: reuse the same RegEx


64J0 (changes requested):
Can you please change the target branch to be the `develop` instead of master? We have this process documented here: https://github.com/giraffe-fsharp/Giraffe?tab=readme-ov-file#tldr.

64J0 (approved):

64J0:
Thanks for the contribution @Thorium!

PR #593, opened 2024-05-04
Change dev instructions and CI

64J0:
## Description

With this PR, I'm proposing a change in the development process, to stop using the `develop` -> `master` branch flow, and start using only the `master` branch. 

* Most people already targets the `master` branch when they are submitting their PRs.

With this in mind, I'm proposing some changes in the README (+ removing the reference to the night builds, since I think they're not running nowadays), DEVGUIDE and the CI workflows.

Related to the CI configuration, I'm separating the build and test workflow, which must work for all PRs and pushes, from the release workflow, which must run only when a new release is created.

Please let me know what you think.

## Related issues

- https://github.com/giraffe-fsharp/giraffe-website/pull/2

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -1,0 +1,41 @@
+name: Build and test
+
+on:
+  push:
+  pull_request:
+
+env:
+  # Stop wasting time caching packages
+  DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
+  # Disable sending usage data to Microsoft
+  DOTNET_CLI_TELEMETRY_OPTOUT: true
+
+# Kill other jobs when we trigger this workflow by sending new commits
+# to the PR.
+# https://stackoverflow.com/a/72408109
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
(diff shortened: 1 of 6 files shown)


TheAngryByrd (dismissed):

nojaf (dismissed):

dbrattli:
Great. Let's use the opportunity to move to a `main` branch instead of `master` which is the default on GitHub these days

64J0:
> Great. Let's use the opportunity to move to a `main` branch instead of `master` since this is the default on GitHub these days

I like this idea @dbrattli. 

Previously I was worried about the impact of this change, for example, the documentation at https://giraffe.wiki/. The contents at this site look pretty much what we have on the markdown files from the `master` branch. But after looking [this code](https://github.com/giraffe-fsharp/giraffe-website/blob/master/src/Giraffe.Website/Program.fs#L253), I think it's going to be easy to update.

Please let me know about other potential impacts that you imagine.

nojaf (dismissed):

dbrattli (approved):
I'm really happy with this change. There might be issues related to renaming branches  🙈 but let's just "fail forward" and fix them if there's any

64J0:
Reverted the commits to avoid replacing the `master` branch with `main` for now.

dbrattli (approved):

Issue #576, opened 2024-02-21
Guid pattern in endpoint router matches invalid values, throws System.FormatException


jcmrva:
Our fuzzer discovered some values that will match the endpoint router's [regex](https://github.com/giraffe-fsharp/Giraffe/blob/6a364955f77609e9265c577ddd817c92fefe104c/src/Giraffe/EndpointRouting.fs#L16) behind `%O`.

App code:
```fsharp
GET [ routef "/try-a-guid/%O" (fun (guid: Guid) -> text $"Success: {guid}" ) ]
```

Test:
```fsharp
#r "nuget: FsHttp"
open FsHttp

let fuzzerInput = "8b3557db-fa-c0c90785ec0b"
http {
    GET $"http://localhost:5000/try-a-guid/{fuzzerInput}"
}
|> Request.send
```

```text
info: Microsoft.AspNetCore.Hosting.Diagnostics[1]
      Request starting HTTP/1.1 GET http://localhost:5000/try-a-guid/8b3557db-fa-c0c90785ec0b - - -
info: Microsoft.AspNetCore.Routing.EndpointMiddleware[0]
      Executing endpoint 'HTTP: GET /try-a-guid/{O0:regex(([0-9A-Fa-f]{{8}}\-[0-9A-Fa-f]{{4}}\-[0-9A-Fa-f]{{4}}\-[0-9A-Fa-f]{{4}}\-[0-9A-Fa-f]{{12}}|[0-9A-Fa-f]{{32}}|[-_0-9A-Za-z]{{22}}))}'
info: Microsoft.AspNetCore.Routing.EndpointMiddleware[1]
      Executed endpoint 'HTTP: GET /try-a-guid/{O0:regex(([0-9A-Fa-f]{{8}}\-[0-9A-Fa-f]{{4}}\-[0-9A-Fa-f]{{4}}\-[0-9A-Fa-f]{{4}}\-[0-9A-Fa-f]{{12}}|[0-9A-Fa-f]{{32}}|[-_0-9A-Za-z]{{22}}))}'
fail: Microsoft.AspNetCore.Server.Kestrel[13]
      Connection id "0HN1J1DK6DFOF", Request id "0HN1J1DK6DFOF:00000001": An unhandled exception was thrown by the application.
      System.FormatException: Unrecognized Guid format.
         at System.Guid.GuidResult.SetFailure(ParseFailure failureKind)
[...]

PR #594, opened 2024-05-10
Fix EndpointRouting Guid regex + tests

64J0:
## Description

With this PR, I'm fixing the EndpointRouting Guid regex ([route constraint](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/routing?view=aspnetcore-8.0#route-constraints)) and adding more automated tests to assert that it's working.

## How to test

You can use this project example:

```fsharp
open System
open Microsoft.AspNetCore.Builder
open Microsoft.Extensions.DependencyInjection
open Microsoft.Extensions.Hosting
open Giraffe
open Giraffe.EndpointRouting

let endpoints =
    [ GET
          [ route "/" (text "Hello World")
            routef "/try-a-guid/%O" (fun (guid: Guid) -> text $"Success: {guid}") ] ]

let notFoundHandler = "Not Found" |> text |> RequestErrors.notFound

let configureApp (appBuilder: IApplicationBuilder) =
    appBuilder.UseRouting().UseGiraffe(endpoints).UseGiraffe(notFoundHandler)

let configureServices (services: IServiceCollection) =
    services.AddRouting().AddGiraffe() |> ignore

[<EntryPoint>]
let main args =
    let builder = WebApplication.CreateBuilder(args)
    configureServices builder.Services

    let app = builder.Build()

    configureApp app
    app.Run()

    0
```

And test with some inputs. For example:

```bash
# must not find
curl http://localhost:5000/try-a-guid/8b3557db-fa-c0c90785ec0b

# must work fine
curl http://localhost:5000/try-a-guid/8b3557db-fa-c0c90785ec
# ...
```

You can find more Guid ideas at the automated tests file.

## Related issues

- Close https://github.com/giraffe-fsharp/Giraffe/issues/576

### DIFF
--- a/src/Giraffe/EndpointRouting.fs
+++ b/src/Giraffe/EndpointRouting.fs
@@ -12,6 +12,10 @@ open FSharp.Core
 open Giraffe
 
 module private RouteTemplateBuilder =
+    // We use this regex route constraint to be compatible with Giraffe's default router,
+    // which supports ShortGuid's.
+    // More information on ASP.NET route constraints:
+    // https://learn.microsoft.com/en-us/aspnet/core/fundamentals/routing?view=aspnetcore-8.0#route-constraints
     let private guidPattern =
         "^[0-9A-Fa-f]{{8}}-[0-9A-Fa-f]{{4}}-[0-9A-Fa-f]{{4}}-[0-9A-Fa-f]{{4}}-[0-9A-Fa-f]{{12}}$|^[0-9A-Fa-f]{{32}}$|^[-_0-9A-Za-z]{{22}}$"
     let private shortIdPattern =
--- a/src/Giraffe/EndpointRouting.fs
+++ b/src/Giraffe/EndpointRouting.fs
@@ -13,7 +13,7 @@ open Giraffe
 
 module private RouteTemplateBuilder =
     let private guidPattern =
-        "([0-9A-Fa-f]{{8}}\-[0-9A-Fa-f]{{4}}\-[0-9A-Fa-f]{{4}}\-[0-9A-Fa-f]{{4}}\-[0-9A-Fa-f]{{12}}|[0-9A-Fa-f]{{32}}|[-_0-9A-Za-z]{{22}})"
+        "^[0-9A-Fa-f]{{8}}-[0-9A-Fa-f]{{4}}-[0-9A-Fa-f]{{4}}-[0-9A-Fa-f]{{4}}-[0-9A-Fa-f]{{12}}$|^[0-9A-Fa-f]{{32}}$|^[-_0-9A-Za-z]{{22}}$"
     let private shortIdPattern =
         "([-_0-9A-Za-z]{{10}}[048AEIMQUYcgkosw])"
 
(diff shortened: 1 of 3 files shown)


src/Giraffe/EndpointRouting.fs
```diff
@@ -13,7 +13,7 @@ open Giraffe
 
 module private RouteTemplateBuilder =
     let private guidPattern =
-        "([0-9A-Fa-f]{{8}}\-[0-9A-Fa-f]{{4}}\-[0-9A-Fa-f]{{4}}\-[0-9A-Fa-f]{{4}}\-[0-9A-Fa-f]{{12}}|[0-9A-Fa-f]{{32}}|[-_0-9A-Za-z]{{22}})"
+        "^[0-9A-Fa-f]{{8}}-[0-9A-Fa-f]{{4}}-[0-9A-Fa-f]{{4}}-[0-9A-Fa-f]{{4}}-[0-9A-Fa-f]{{12}}$|^[0-9A-Fa-f]{{32}}$|^[-_0-9A-Za-z]{{22}}$"
```
dbrattli, line 16: The last part `|^[-_0-9A-Za-z]{{22}}$` matches for a 22 char GUID e.g `0000000000000000000000` but there´s no test for it. Is there a format for such a GUID?

64J0: It's this [ShortGuid](https://github.com/giraffe-fsharp/Giraffe/blob/master/src/Giraffe/ShortGuid.fs) format. I'll add a test for it.

64J0: Done with this commit https://github.com/giraffe-fsharp/Giraffe/pull/594/commits/b57fac833b4cb78b9524f1c9a1ee267d765e6ac2

nojaf, line 16: Maybe add a comment on why this is done with Regex in the first place.
Without much context, I would assume Guid.TryParse would do the trick here.

dbrattli (dismissed):
LGTM!

src/Giraffe/EndpointRouting.fs
64J0: Indeed, my initial approach was to try to use the `Guid.TryParse`, but was not able. If this endpoint worked with normal Guid, it would only require a [route constraint](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/routing?view=aspnetcore-8.0#route-constraints) like `{id:guid}`, but in order to be compatible with Giraffe custom router we must support [ShortGuid](https://github.com/giraffe-fsharp/Giraffe/blob/master/src/Giraffe/ShortGuid.fs) too.

64J0: Added the comment with this commit https://github.com/giraffe-fsharp/Giraffe/pull/594/commits/d7188c7aa44e6688516fa74566876a2646cd3b4b

64J0: This same regex (+/-) is used here https://github.com/giraffe-fsharp/Giraffe/blob/master/src/Giraffe/FormatExpressions.fs#L30-L31

dbrattli (approved):

nojaf (approved):
Great job!

PR #596, opened 2024-05-13
Fix pre-release/release workflows

64J0:
## Description

Although the `prerelease` workflow was running with the latest PR changes, it was still not creating the new nuget version, since it was not running the [dotnet nuget push](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-nuget-push) command.

After realizing this problem, I figured out that in this case it would be better to use the `deploy` workflow that does the job correctly. This way, I'm proposing the removal of `prerelease` workflow and the usage of `deploy` for both `pre-release`s and `release`s on GitHub.

Please let me know what you think.

## How to test

Check this CI execution: [link](https://github.com/giraffe-fsharp/Giraffe/actions/runs/9070910047/job/24923641652?pr=596).

Notice that the "new" pre-release version was created and it's available on [nuget](https://www.nuget.org/packages/Giraffe#versions-body-tab):

image

## Related issues

- Continuation of https://github.com/giraffe-fsharp/Giraffe/pull/595

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,6 +1,7 @@
 name: Publish
 
 on:
+  push:
   release:
     types:
       - published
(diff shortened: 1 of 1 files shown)


dbrattli (approved):
Looks like a good idea to simplify this. Let's try it!

TheAngryByrd:
👋 Just be aware of prerelease versioning scheme is semantic versioning. Meaning if you release `v6.4.1-alpha-3` then eventually `v6.4.1-alpha-10` that the previous version is ranked higher than the later version because "3" comes before "10".

baronfel:
Jimmy's right - what you actually want is a scheme like `6.4.1-alpha.<INCREMENTING_VERSION>` because per SemVer each _section_ of the prerelease part (e.g. `alpha.<VERSION>`) is compared separately, and purely-numeric sections are compared as integers.

This is why you see versions of the .NET SDK like `9.0.100-preview.3`

64J0:
Got it, thanks for the heads-up @TheAngryByrd and @baronfel. I'll keep this in mind for next versions, and will add to the DEVELOP document to keep it recorded.

PR #600, opened 2024-07-02
Tiny memory optimization

Thorium:
CA1835: Prefer the memory-based overloads of ReadAsync/WriteAsync methods in stream-based classes

https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1835

### DIFF
--- a/src/Giraffe/HttpContextExtensions.fs
+++ b/src/Giraffe/HttpContextExtensions.fs
@@ -367,7 +367,7 @@ type HttpContextExtensions() =
                 let contentLength = if is205StatusCode then 0 else bytes.Length
                 ctx.SetHttpHeader(HeaderNames.ContentLength, contentLength)
             if ctx.Request.Method <> HttpMethods.Head then
-                do! ctx.Response.Body.WriteAsync(bytes, 0, bytes.Length)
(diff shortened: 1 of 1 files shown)


64J0:
Thanks for opening this PR @Thorium! 

I tried to benchmark how this change would improve the project and the code is available at this other PR (https://github.com/giraffe-fsharp/Giraffe/pull/601), but was not able to get a consistent result. Probably due to some misconfiguration there. The results from the last execution are mentioned in this PR description.

Thorium:
I would expect that if you don't do AsMemory, then it internally copies the byte-array so it's twice in memory. Which you don't notice if your byte[] is size of 1kB, but if your byte[] is e.g. 2GB then it should really matter.

Thorium:
Ok it seems to do it internally anyway
https://github.com/dotnet/aspnetcore/blob/98c4546ddccfe2f62c4dc239fabd0c92848b255d/src/Servers/Kestrel/Core/src/Internal/Http/HttpResponseStream.cs#L92-L95
...so this is not needed

64J0:
Got it, thanks for the help to understand it.

Issue #564, opened 2023-10-26
Never decalre reader with `use` on `ctx.Request.Body`


xperiandri:
this
https://github.com/giraffe-fsharp/Giraffe/blob/6a364955f77609e9265c577ddd817c92fefe104c/src/Giraffe/HttpContextExtensions.fs#L197
disposes body, because the reader disposes underlying stream.
Which must not be done in any case!

64J0:
I did not get it, what is the problem with disposing the request body after reading it?

xperiandri:
The problem is that Giraffe is not an owner of it. The owner of it is an appropriate ASP.NET core middleware.
Giraffe must not dispose the stuff it is not owner of.
It leads to unexpected behavior for the consumer.

64J0:
I see, thanks for the explanation. I agree then!

Banashek:
Looks like the `leaveOpen` parameter could be specified to allow disposal of the `StreamReader` without disposing the underlying resource: [doc reference](https://learn.microsoft.com/en-us/dotnet/api/system.io.streamreader.-ctor?view=net-8.0#system-io-streamreader-ctor(system-io-stream-system-text-encoding-system-boolean-system-int32-system-boolean))

```fs
use reader = new StreamReader(ctx.Request.Body, Encoding.UTF8, leaveOpen = true)
```

PR #615, opened 2024-09-06
Fix `ReadBodyFromRequestAsync` disposing `ctx.Request.Body`

64J0:
## Description

As noticed by @xperiandri, the `ReadBodyFromRequestAsync` seems to be incorrectly disposing the `ctx.Request.Body`. I'm adopting the suggestion from @Banashek, to use the `leaveOpen = true` parameter.

From the [docs](https://learn.microsoft.com/en-us/dotnet/api/system.io.streamreader.-ctor?view=net-8.0#overloads):

> Unless you set the leaveOpen parameter to true, the [StreamReader](https://learn.microsoft.com/en-us/dotnet/api/system.io.streamreader?view=net-8.0) object calls [Dispose()](https://learn.microsoft.com/en-us/dotnet/api/system.io.stream.dispose?view=net-8.0#system-io-stream-dispose) on the provided [Stream](https://learn.microsoft.com/en-us/dotnet/api/system.io.stream?view=net-8.0) object when [StreamReader.Dispose](https://learn.microsoft.com/en-us/dotnet/api/system.io.streamreader.dispose?view=net-8.0) is called.

## How to test

Check the automated tests.

*Thanks @Banashek and @1eyewonder for the comments and ideas.

## Related issues

- Close https://github.com/giraffe-fsharp/Giraffe/issues/564

### DIFF
--- a/tests/Giraffe.Tests/Helpers.fs
+++ b/tests/Giraffe.Tests/Helpers.fs
@@ -196,9 +196,14 @@ let hasLastModified (lastModified: DateTimeOffset) (response: HttpResponseMessag
     Assert.Equal(lastModified, response.Content.Headers.LastModified.Value)
     response
 
+let getReqBody (ctx: HttpContext) =
+    ctx.Request.Body.Position <- 0L
+    use reader = new StreamReader(ctx.Request.Body, Encoding.UTF8, leaveOpen = true)
+    reader.ReadToEnd()
+
 let getBody (ctx: HttpContext) =
     ctx.Response.Body.Position <- 0L
-    use reader = new StreamReader(ctx.Response.Body, Encoding.UTF8)
+    use reader = new StreamReader(ctx.Response.Body, Encoding.UTF8, leaveOpen = true)
     reader.ReadToEnd()
 
 let readText (response: HttpResponseMessage) = response.Content.ReadAsStringAsync()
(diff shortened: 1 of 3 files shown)


64J0:
I'm not sure how to test this properly. Ideas are welcome!

Banashek:
The original bug report was that the body gets disposed.

I would think that asserting the presence of the body post-`use` would be a good test.

nojaf (approved):

1eyewonder:
@64J0 Here is a similar type of scenario I had to add over in FsToolkit.ErrorHandling to deal with dispose issues https://github.com/demystifyfp/FsToolkit.ErrorHandling/pull/271 if you were still looking for some unit test ideas

64J0:
Thanks for the ideas folks! I'll work on it during this week

tests/Giraffe.Tests/Helpers.fs
```diff
@@ -196,9 +196,14 @@ let hasLastModified (lastModified: DateTimeOffset) (response: HttpResponseMessag
     Assert.Equal(lastModified, response.Content.Headers.LastModified.Value)
     response
 
+let getReqBody (ctx: HttpContext) =
```
nojaf, line 199: what is the difference with `getBody`?

nojaf (approved):

tests/Giraffe.Tests/Helpers.fs
64J0: One reads from the Response, while the other reads from the Request

nojaf: Okay

Issue #618, opened 2024-10-02
Asp.NET Rate Limiter middleware support


GallaFrancesco:
https://learn.microsoft.com/en-us/aspnet/core/performance/rate-limit?view=aspnetcore-8.0 is a rate limiting middleware implementation and I'm struggling to make this work in Giraffe. Two main reasons:

1. I couldn't find a way to enable the rate limiter on specific endpoints using `RequireRateLimiting()` in a similar way as this sample Asp.NET project: https://github.com/dotnet/AspNetCore.Docs.Samples/blob/23fbafb68257b59bf6f3775fe12c51afc2f242e3/fundamentals/middleware/rate-limit/WebRateLimitAuth/Program.cs#L163

It would be ideal to support a middleware such as:
```fsharp
route "/limited" >=> text "This is rate limited" >=> requireRateLimiting policy
```
Though I could not find any example of this.

2. I find it difficult to configure the middleware from F#, but I might be doing something wrong or excessively complicated. `AddTokenBucketLimiter` is failing when trying to call it from F#, with ` error FS0193: Type constraint mismatch. The type     'RateLimiterOptions'    is not compatible with type    'unit'`, see the following snippet. 

```fsharp
type SampleRateLimitOptions =
    { ReplenishmentPeriod: int
      QueueProcessingOrder: QueueProcessingOrder
      QueueLimit: int 
      TokenLimit: int 
      TokensPerPeriod: int 
      AutoReplenishment: bool }

    static member Default : SampleRateLimitOptions = // times in seconds
        { ReplenishmentPeriod = 1
          QueueProcessingOrder = QueueProcessingOrder.OldestFirst
          QueueLimit = 10
          TokenLimit = 10
          TokensPerPeriod = 4
          AutoReplenishment = true }

let configureRateLimiter (builder: WebApplicationBuilder) =
    let rateLimitOptions = SampleRateLimitOptions.Default 

    let loadRateLimiterOptions (options: TokenBucketRateLimiterOptions) =
        options.TokenLimit <- rateLimitOptions.TokenLimit
        options.QueueProcessingOrder <- QueueProcessingOrder.OldestFirst
        options.QueueLimit <- rateLimitOptions.QueueLimit
        options.ReplenishmentPeriod <- TimeSpan.FromSeconds(rateLimitOptions.ReplenishmentPeriod)
        options.TokensPerPeriod <- rateLimitOptions.TokensPerPeriod
[...]

PR #619, opened 2024-10-05
Add global rate limiting sample and much more

64J0:
## Description

> [!WARNING]
> This is just experimental. To be honest I don't like this configuration that demands an empty list. I'll ask for more opinions before making this huge change to the Giraffe codebase.

With this PR, I'm:

* Adding a global rate limiting sample;
* Updating the code to work only with .NET 8 (due to the ASP.NET extensions);
* Adding the possibility to configure "ASP.NET extensions", which are middlewares like [CacheOutput](https://learn.microsoft.com/en-us/aspnet/core/performance/caching/output?view=aspnetcore-8.0) and [RateLimiting](https://learn.microsoft.com/en-us/aspnet/core/performance/rate-limit?view=aspnetcore-8.0), per route;
* Adding a new rate limiting sample that deals with this mechanism.

TODO:

- [ ] Filter repeated "ASP.NET extensions" to use the last value only;
- [ ] Update docs;
- [ ] Q: Is it possible to add OpenApi native configuration from ASP.NET?;
- [ ] Add automated tests.

## How to test

Check the instructions at the sample's README.md.

## Related issues

Related to https://github.com/giraffe-fsharp/Giraffe/issues/618.


src/Giraffe/EndpointRouting.fs
```diff
@@ -232,39 +270,48 @@ module Routers =
     let TRACE = applyHttpVerbToEndpoints TRACE
     let CONNECT = applyHttpVerbToEndpoints CONNECT
 
-    let route (path: string) (handler: HttpHandler) : Endpoint =
-        SimpleEndpoint(HttpVerb.NotSpecified, path, handler, id)
+    let route (path: string) (aspNetExtensions: AspNetExtension list) (handler: HttpHandler) : Endpoint =
```
nojaf, line 273: This is massive breaking change is it not?
Is means every existing code with `route` would break.

I'd go for the following:
```fsharp
let routeWithExtensions (path: string) (aspNetExtensions: AspNetExtension list) (handler: HttpHandler) : Endpoint = SimpleEndpoint(HttpVerb.NotSpecified, path, handler, id, aspNetExtensions)

let route (path: string) (handler: HttpHandler) : Endpoint = routeWithExtensions(path, [], handler)
```

src/Giraffe/Giraffe.fsproj
```diff
@@ -8,7 +8,7 @@
     <NeutralLanguage>en-GB</NeutralLanguage>
 
     <!-- Build settings -->
-    <TargetFrameworks>net6.0;net7.0;net8.0</TargetFrameworks>
+    <TargetFrameworks>net8.0</TargetFrameworks>
```
nojaf, line 11: Another breaking change here.

src/Giraffe/Giraffe.fsproj
```diff
@@ -54,16 +54,16 @@
     <IncludeAssets>build</IncludeAssets>
     <PrivateAssets>all</PrivateAssets>
     </PackageReference>
-    <PackageReference Include="G-Research.FSharp.Analyzers" Version="0.10.0">
+    <PackageReference Include="G-Research.FSharp.Analyzers" Version="0.11.0">
```
nojaf, line 57: Updating analyzers is best handled in a separate PR, to be honest. It can make the current PR noisy, and I generally find that smaller PRs focusing on one thing are easier to review.

nojaf (changes requested):
One user on https://github.com/giraffe-fsharp/Giraffe/issues/618 requesting a feature is typically not a strong enough reason to make drastic changes to the public API. I would only consider it if the demand were significantly higher.

As a workaround, you can use:

```fsharp
    type Endpoint =
        | SimpleEndpoint of HttpVerb * RouteTemplate * HttpHandler * ConfigureEndpoint
        | TemplateEndpoint of
            HttpVerb *
            RouteTemplate *
            RouteTemplateMappings *
            (obj -> HttpHandler) *
            ConfigureEndpoint
        | NestedEndpoint of RouteTemplate * Endpoint list * ConfigureEndpoint
        | MultiEndpoint of Endpoint list
#if NET8_0
        | NewThingWithRateLimiting of blah
#endif
```

Additionally, consider introducing a similar DSL function wrapped in `#if NET8_0`. I understand this may complicate things, but it's a better approach than altering the API on a whim.

Alternatively, you could release another major version, but be cautious not to have too many, as end-users will always have upgrade work on their side.

On a practical note, launching another major version to drop support for net6.0 after the release of net9.0 in November seems reasonable. Ultimately, I believe this change should be communicated more clearly and not be a side effect of a PR.

src/Giraffe/EndpointRouting.fs
64J0: Indeed, it looks a better path to go

ebellani:
:+1:

64J0:
I'm closing this PR in favor of more succinct ones, which I'll open later, tackling specific points of the ideas presented here.

src/Giraffe/EndpointRouting.fs
```diff
@@ -232,39 +270,48 @@ module Routers =
...
+        SimpleEndpoint(HttpVerb.NotSpecified, path, handler, id, aspNetExtensions)
```
64J0, line 274: Something that has potential 🤔 

This function that currently uses the [id](https://fsharp.github.io/fsharp-core-docs/reference/fsharp-core-operators.html#id) function has this type:

```fsharp
type ConfigureEndpoint = IEndpointConventionBuilder -> IEndpointConventionBuilder
```

I don't know much about it, but maybe we can use it instead of this custom wrapper for Asp.Net extensions https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.builder.iendpointconventionbuilder?view=aspnetcore-9.0

Issue #521, opened 2022-08-27
Assembly version numbers have reverted to default 1.0.0.0 ever since NuGet version 5.0.0-alpha-002


DalekBaldwin:
I haven't noticed this causing any build issues yet, but it is tripping up some other development tools.

64J0:
Hi @DalekBaldwin, thanks for opening this issue. Do you know if it's still happening nowadays?

DalekBaldwin:
Yes, it's still happening. You can inspect the metadata from here: https://nuget.info/packages/Giraffe/6.4.0

64J0:
I noticed that this happens for other projects from the F# community as well. For example [SQLProvider](https://nuget.info/packages/SQLProvider/1.4.0):

![image](https://github.com/user-attachments/assets/f61837ea-dd04-44ac-80f0-c13e2e6198b9)

The fix is apparently to have this [AssemblyInfo](https://github.com/fsprojects/SQLProvider/blob/master/src/Common/AssemblyInfo.fs) file defining the values. I found it on [Fable](https://github.com/fable-compiler/Fable/blob/main/src/Fable.Core/AssemblyInfo.fs) too.

baronfel:
Yous shouldn't need a dedicated assembly info file FYI - merely setting the Version property in your project file is enough to ensure that all of the relevant versioning data is stamped into your assembly.

nojaf:
Or use https://github.com/ionide/KeepAChangelog

64J0:
> Or use https://github.com/ionide/KeepAChangelog

Related https://github.com/ionide/KeepAChangelog/issues/34

PR #629, opened 2024-11-17
Add AssemblyVersion attribute

64J0:
## Description

With this PR, I'm adding the **AssemblyVersion** attribute to Giraffe.fsproj, and updating the DEVGUIDE document, adding a new step to update this attribute whenever a new release is intended.

## Related issues

- Close https://github.com/giraffe-fsharp/Giraffe/issues/521

### DIFF
--- a/src/Giraffe/Giraffe.fsproj
+++ b/src/Giraffe/Giraffe.fsproj
@@ -2,6 +2,7 @@
   <PropertyGroup>
     <!-- General -->
     <AssemblyName>Giraffe</AssemblyName>
+    <AssemblyVersion>7.0.2</AssemblyVersion>
     <Description>A native functional ASP.NET Core web framework for F# developers.</Description>
     <Copyright>Copyright 2020 Dustin Moris Gorski</Copyright>
     <Authors>Dustin Moris Gorski and contributors</Authors>
--- a/DEVGUIDE.md
+++ b/DEVGUIDE.md
@@ -8,23 +8,25 @@ This documentation must be used as a guide for maintainers and developers for bu
     1. `git checkout master`
 2. Add a new entry at the top of the RELEASE_NOTES.md with a version and a date.
     1. If possible link to the relevant issues and PRs and credit the author of the PRs
-3. Create a new commit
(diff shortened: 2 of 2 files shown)


64J0:
@baronfel and @DalekBaldwin, if you have some free time, please take a look at this PR. I think this must be enough to close the issue https://github.com/giraffe-fsharp/Giraffe/issues/521, and later we can revisit this process, to automate it using https://github.com/ionide/KeepAChangelog.

src/Giraffe/Giraffe.fsproj
```diff
@@ -2,6 +2,7 @@
   <PropertyGroup>
     <!-- General -->
     <AssemblyName>Giraffe</AssemblyName>
+    <AssemblyVersion>7.0.2</AssemblyVersion>
```
64J0, line 5: Another potential fix would be to define this property at the `dotnet pack` command on CI here https://github.com/giraffe-fsharp/Giraffe/blob/master/.github/workflows/publish.yml#L69

nojaf (approved):

mrtz-j (approved):

kerams:
Is this causing the assembly file version to be `0.0.0`? The previous releases used `1.0.0`, so Fable.Remoting now throws an exception when used with the v8 alpha.

64J0:
> Is this causing the assembly file version to be `0.0.0`? The previous releases used `1.0.0`, so Fable.Remoting now throws an exception when used with the v8 alpha.

It must be related. I just noticed that although the `AssemblyFileVersion` is defined properly for the alpha release (https://nuget.info/packages/Giraffe/8.0.0-alpha-001):

image

The version value at the Full Name is not.

This is what we had before:

image

64J0:
I think this must be enough to fix this problem @kerams: https://github.com/giraffe-fsharp/Giraffe/pull/655. Once this is fixed, I'm going to generate a new alpha release.

64J0:
Now it's showing something reasonable https://nuget.info/packages/Giraffe/8.0.0-alpha-002.

image

kerams:
Yeah, it works, thank you.

PR #639, opened 2025-01-16
Add .NET 9 as target framework, fine-tune dependabot, update CI and clean tests removing .NET 6/7 from target frameworks

64J0:
## Description

With this PR, 

- I'm adding .NET 9 as a target framework;
- Changing **fantomas** CI to run using ubuntu-latest;
- Changing **build-and-test** push trigger to execute only for pushes to the `master` branch;
- Update dependabot configuration to scan multiple [directories](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#directories-or-directory--);
- Clean test dependencies and remove .NET 6 and 7 from target frameworks.

## How to test

Check CI.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -2,6 +2,8 @@ name: Build and test
 
 on:
   push:
+    branches:
+      - master
   pull_request:
 
 env:
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -33,6 +33,7 @@ jobs:
             6.x
             7.x
             8.x
+            9.x
       - name: Restore
         run: dotnet restore
       - name: Build
@@ -48,7 +49,7 @@ jobs:
       - name: Setup .NET Core
         uses: actions/setup-dotnet@v4
         with:
-          dotnet-version: 8.x
+          dotnet-version: 9.x
       - name: Restore tools
         run: dotnet tool restore
       - name: Build solution
--- a/.github/workflows/fantomas-check.yml
+++ b/.github/workflows/fantomas-check.yml
@@ -14,7 +14,7 @@ concurrency:
 jobs:
   fantomas-check:
     name: Code format check
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
--- a/.github/workflows/fantomas-check.yml
+++ b/.github/workflows/fantomas-check.yml
@@ -14,7 +14,7 @@ concurrency:
 jobs:
   fantomas-check:
     name: Code format check
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-latest
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
--- a/tests/Giraffe.Tests/Giraffe.Tests.fsproj
+++ b/tests/Giraffe.Tests/Giraffe.Tests.fsproj
@@ -1,6 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>net6.0;net7.0;net8.0</TargetFrameworks>
+    <TargetFrameworks>net6.0;net7.0;net8.0;net9.0</TargetFrameworks>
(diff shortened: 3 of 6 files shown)


TheAngryByrd (approved):

.github/workflows/fantomas-check.yml
```diff
@@ -14,7 +14,7 @@ concurrency:
 jobs:
   fantomas-check:
     name: Code format check
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
```
nojaf, line 17: Would go with ubuntu-latest here really.

.github/workflows/build-and-test.yml
```diff
@@ -33,6 +33,7 @@ jobs:
             6.x
             7.x
             8.x
+            9.x
```
nojaf, line 36: It it a bit weird that you install all these framework, to not have them in a matrix.

tests/Giraffe.Tests/Giraffe.Tests.fsproj
```diff
@@ -1,6 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>net6.0;net7.0;net8.0</TargetFrameworks>
+    <TargetFrameworks>net6.0;net7.0;net8.0;net9.0</TargetFrameworks>
```
nojaf, line 3: Do you aim to remove net6 and net7 in a different PR?
Feels weird to keep them.

TheAngryByrd: Yeah, I think it makes sense to remove net6/7 from testing.

I really want Microsoft to have clear documentation on when it makes sense to update a TFM. Because I don't think there's a super compelling reason to build Giraffe project for 9 if we're not doing any ifdefs.

.github/workflows/fantomas-check.yml
64J0: Done with https://github.com/giraffe-fsharp/Giraffe/pull/639/commits/85b8c9019dd79b5ccb12cbbdf5344952d2a52550

tests/Giraffe.Tests/Giraffe.Tests.fsproj
64J0: I get this warning when building this project locally:

```
ResponseCachingApp succeeded with 1 warning(s) (2.4s) → samples/ResponseCachingApp/bin/Debug/net7.0/ResponseCachingApp.dll
    /.../sdk/9.0.101/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.EolTargetFrameworks.targets(32,5): warning NETSDK1138: The target framework 'net7.0' is out of support and will not receive security updates in the future. Please refer to https://aka.ms/dotnet-core-support for more information about the support policy.
```

.github/workflows/build-and-test.yml
64J0: I think that for this specific workflow, it makes more sense to have all the used .NET versions available at the same runner process. Other than this, it would trigger too many jobs. Maybe we can test this approach in another PR, what do you think?

tests/Giraffe.Tests/Giraffe.Tests.fsproj
64J0: Removed with this commit https://github.com/giraffe-fsharp/Giraffe/pull/639/commits/0cfba09cd534fe97dcd9e075a671735dc18cf8c0

.github/workflows/build-and-test.yml
nojaf: If you care to support the combination of all these dotnet versions on all these operating systems, then a matrix is the way to go.
I wouldn't worry about too many jobs as these will run in parallel anyway.

nojaf (approved):

.github/workflows/build-and-test.yml
64J0: Indeed, it may be better to get specific problems. I'll open a PR testing this setup later today

mrtz-j (approved):

.github/workflows/build-and-test.yml
64J0: https://github.com/giraffe-fsharp/Giraffe/pull/644

PR #644, opened 2025-01-20
Use matrix for build-and-test CI workflow

64J0:
## Description

Use a matrix strategy for build-and-test CI workflow.

## How to test

Check CI.

## Related issues

Continuation of https://github.com/giraffe-fsharp/Giraffe/pull/639.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -25,17 +25,14 @@ jobs:
     strategy:
       matrix:
         os: [ ubuntu-latest, windows-latest, macos-latest ]
+        version: [ 6.x, 7.x, 8.x, 9.x ]
     steps:
       - name: Checkout
         uses: actions/checkout@v4
       - name: Setup .NET Core
         uses: actions/setup-dotnet@v4
         with:
-          dotnet-version: |
-            6.x
-            7.x
-            8.x
-            9.x
+          dotnet-version: ${{ matrix.version }}
       - name: Restore
         run: dotnet restore
       - name: Build


TheAngryByrd:
I'm actually not sure why a matrix would be needed. The test project already tests against different runtimes.

64J0:
> I'm actually not sure why a matrix would be needed. The test project already tests against different runtimes.

From my understanding, the idea is to test each version alone in a single environment. It's related to this discussion: https://github.com/giraffe-fsharp/Giraffe/pull/639#discussion_r1919660153.

Anyway, this is currently not feasible due to the samples, as shown by this PR, since they demand multiple .NET versions to build.

TheAngryByrd:
Yeah, I don't understand @nojaf's original comment. Unless you have to do it because of weird tooling testing, doesn't make sense to me to put it in a matrix. I guess you could run a TFM specifically in a runner but seems like overkill if the tests are fast enough.

64J0:
I'll close this PR for now, later we can revisit this decision.

nojaf:
My remark was more about running the sample with that specific SDK installed.
Right now, you install all of them but that might not be case on the user's machine.

64J0:
It must be feasible when targeting Giraffe's `src/` folder. What do you think about adding a new workflow that runs with a matrix strategy just to check this folder's codebase using specific OS + SDK combinations?

nojaf:
Well, it is probably doable but I don't really have any opinion on this.
Might be overkill and you could just install latest 9 SDK.
I'm not sure what it really changes in the grand scheme of things.

Issue #611, opened 2024-08-11
Naming parameters in endpoint path


pir2o61:
It seems that there is no way to name parameters in path.
Something like {petId:%i} in https://github.com/hvester/EndpointBuilder will be very nice.

On way that it can be done is via changing `convertToRouteTemplate` function in `EndpointRouting`. Something like below should be enough:
```
            | '%' :: c :: tail ->
                let (placeholderName, placeholderTemplate), newTail = 
                    match tail with
                    | ':' :: stail ->
                        let splitIndex = stail |> List.tryFindIndex (fun c -> c = '/')
                        match splitIndex with
                        | Some splitIndex ->
                            let name , newTail = 
                                stail 
                                |> List.splitAt splitIndex
                            getConstraint i c (Some (System.String.Concat(Array.ofList(name)))), newTail
                        | None -> 
                            getConstraint i c (Some (System.String.Concat(Array.ofList(stail)))), []
                    | _ ->
                        getConstraint i c None, tail
                let template, mappings = convert (i + 1) newTail
                placeholderTemplate + template, (placeholderName, c) :: mappings
```

where `getConstraint` is changed like this
```
    let private getConstraint (i : int) (c : char) (name : string option) =
        let name = Option.defaultValue (sprintf "%c%i" c i) name
```

1. Do we want this feature?
2. On what scope it should work? (what names do we allow, do we want to have it in format `%c:name`. `name:%c' or different

RJSonnenberg:
Yes. We want this feature. Otherwise, the Giraffe.OpenApi feels kind of incomplete since the path parameters are just interpreted as `i1`, `s2`, etc.

I'm assuming this feature is related to https://github.com/giraffe-fsharp/Giraffe.OpenApi/issues/6 ?

64J0:
Do you think we can close this issue after [this PR](https://github.com/giraffe-fsharp/Giraffe/pull/656) is merged?

RJSonnenberg:
I think this can be closed now. @pir2o61 , what do you think?

64J0:
Closing this issue, since there were no new interactions for a while, and I think that the PRs mentioned before already cover the goal of it. Feel free to open a new issue later if this is not the case.

PR #656, opened 2025-05-22
Enhance routef support for named parameters and improve documentation

RJSonnenberg:
## Description

Add support for named parameters to `routef`. This allows you to use the syntax `%i:barcode` to name a parameter.

This will allow the Giraffe.OpenApi package to have named parameters instead of their names being simple `[type][index]` (e.g., "i0").

## How to test

1. Pull down the repo.
2. Run the EndpointRoutingApp sample app.
3. Navigate to `https://localhost:56830/pet/123`.
4. Should see text: "PetId: 123"

## Related issues

https://github.com/giraffe-fsharp/Giraffe/issues/611

### DIFF
--- a/tests/Giraffe.Tests/EndpointRoutingTests.fs
+++ b/tests/Giraffe.Tests/EndpointRoutingTests.fs
@@ -154,11 +154,48 @@ let ``subRouteWithExtensions: GET request returns expected result`` (path: strin
 
 [<Theory>]
 [<InlineData("/pet/42", "PetId: 42")>]
+[<InlineData("/pet/0", "PetId: 0")>]
+[<InlineData("/pet/123", "PetId: 123")>]
+[<InlineData("/pet/-1", "PetId: -1")>]
+[<InlineData("/pet/abc", "Not Found")>]
+[<InlineData("/pet/123abc", "Not Found")>]
+[<InlineData("/pet/123.456", "Not Found")>]
+[<InlineData("/pet/123-456", "Not Found")>]
 let ``routef: GET "/pet/%i:petId" returns named parameter`` (path: string, expected: string) =
     task {
         let endpoints: Endpoint list =
             [
                 GET [ routef "/pet/%i:petId" (fun (petId: int) -> text ($"PetId: {petId}")) ]
+                GET [ routef "/foo/%i/bar/%i:barId" (fun (fooId: int, barId: int) -> text ($"FooId: {fooId}, BarId: {barId}")) ]
+            ]
+
+        let notFoundHandler = "Not Found" |> text |> RequestErrors.notFound
+
+        let configureApp (app: IApplicationBuilder) =
+            app.UseRouting().UseGiraffe(endpoints).UseGiraffe(notFoundHandler)
+
+        let configureServices (services: IServiceCollection) =
+            services.AddRouting().AddGiraffe() |> ignore
+
+        let request = createRequest HttpMethod.Get path
+
+        let! response = makeRequest (fun () -> configureApp) configureServices () request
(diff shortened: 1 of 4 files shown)


64J0:
Thanks for opening this PR @RJSonnenberg. I'll review it either this or next week.

Btw, I enabled the CI workflows to run for this PR, and noticed that `fantomas` is breaking. Can you please fix this?

You can use the following commands on your terminal:

```bash
# considering that you're at the root directory of this project

dotnet tool restore

dotnet fantomas .
```

tests/Giraffe.Tests/EndpointRoutingTests.fs
```diff
@@ -151,3 +151,27 @@ let ``subRouteWithExtensions: GET request returns expected result`` (path: strin
 
         content |> shouldEqual expected
     }
+
+[<Theory>]
+[<InlineData("/pet/42", "PetId: 42")>]
+let ``routef: GET "/pet/%i:petId" returns named parameter`` (path: string, expected: string) =
```
64J0, line 157: Some additional points I'd like to cover with our automated tests:

- Mix of named and unnamed parameters, just to assert this combination works;
- A new test to clarify what happens if the informed data type is not compatible with the expected type (for example, this `/pet` endpoint expects an integer, I'd like to have a test asserting what happens if we send a string, or a float);
- Explore what happens when using named parameters and the handler presents this in an unordered way. For example:
  ```fsharp
  // will this work? was it supposed to work?
  let handlerNamedUnordered (riverId: int, petId: int) : HttpHandler =
    fun (_: HttpFunc) (ctx: HttpContext) -> sprintf "PetId: %i, RiverId: %i" petId riverId |> ctx.WriteTextAsync

  GET [
    // ...
    routef "/pet/%i:petId/river/%i:riverId" handlerNamedUnordered
  ]
  ```

P.S. Notice that for this last item, I'm considering the changes you made to the `EndpointRoutingApp/Program.fs` file.

RJSonnenberg: I'll try to address these in a timely manner. Thank you for the feedback!

64J0: Nice, thanks for adding these new tests. I'll finish reviewing this PR during this week.

tests/Giraffe.Tests/EndpointRoutingTests.fs
```diff
@@ -151,3 +151,72 @@ let ``subRouteWithExtensions: GET request returns expected result`` (path: strin
...
+        let! content = response |> readText
+        content |> shouldEqual expected
+    }
+
+[<Theory>]
+[<InlineData("/foo/123/bar/abc", "FooId: 123, BarId: abc")>]
+[<InlineData("/foo/999/bar/789", "FooId: 999, BarId: 789")>]
+[<InlineData("/foo/-1/bar/123", "FooId: -1, BarId: 123")>]
+[<InlineData("/foo/abc/bar/def", "Not Found")>]
+let ``routef: GET "/foo/%i:fooId/bar/%i/baz/%s:bazId" returns named and unnamed parameters``
```
64J0, line 196: I think this is a more correct name for this test, since we don't check the `/baz` part.

```suggestion
let ``routef: GET "/foo/%i:fooId/bar/%i" returns named and unnamed parameters``
```

64J0 (approved):
This change looks good to me. 
I'll wait until this Wednesday before merging, so we give more time for other maintainers to also take a look at it.

nojaf:
I don't need more time, I'm not suited to review this one.
Best if others chime in here.

PR #692, opened 2025-11-02
Could we avoid allocation of UTF8 byte array?

Thorium:
Could we avoid allocation of UTF8 byte array?

---

Another code-fixes to consider as well:
https://github.com/giraffe-fsharp/Giraffe/blob/c00ace4185277ba96f90b3cc617a434f75999161/src/Giraffe/Helpers.fs#L10
The RecyclableMemoryStreamManager could be constructed like this when called:

```fsharp
let internal recyclableMemoryStreamManager = Lazy<RecyclableMemoryStreamManager>(fun () -> RecyclableMemoryStreamManager())
```
...and then it wouldn't be null ever.

---

I'm not sure if the portMap dictionary creation and filling should be inside or outside of the `fun next` here:
https://github.com/giraffe-fsharp/Giraffe/blob/c00ace4185277ba96f90b3cc617a434f75999161/src/Giraffe/Routing.fs#L54
...because now it constructs a new dictionary on each next-call. Will there be more next-calls than routePorts calls? Just to note that Dictionary is not thread-safe so not to put it outside `let routePorts`.

### DIFF
--- a/src/Giraffe/Core.fs
+++ b/src/Giraffe/Core.fs
@@ -503,8 +503,7 @@ module Core =
     /// <param name="str">The string value to be send back to the client.</param>
     /// <returns>A Giraffe <see cref="HttpHandler" /> function which can be composed into a bigger web application.</returns>
     let setBodyFromString (str: string) : HttpHandler =
-        let bytes = Encoding.UTF8.GetBytes str
-        fun (_: HttpFunc) (ctx: HttpContext) -> ctx.WriteBytesAsync bytes
+        fun (_: HttpFunc) (ctx: HttpContext) -> ctx.WriteStringAsync str
 
     /// <summary>
     /// Writes an UTF-8 encoded string to the body of the HTTP response and sets the HTTP Content-Length header accordingly, as well as the Content-Type header to text/plain.


src/Giraffe/Core.fs
```diff
@@ -435,8 +435,7 @@ module Core =
     /// <param name="str">The string value to be send back to the client.</param>
     /// <returns>A Giraffe <see cref="HttpHandler" /> function which can be composed into a bigger web application.</returns>
     let setBodyFromString (str: string) : HttpHandler =
-        let bytes = Encoding.UTF8.GetBytes str
-        fun (_: HttpFunc) (ctx: HttpContext) -> ctx.WriteBytesAsync bytes
+        fun (_: HttpFunc) (ctx: HttpContext) -> ctx.WriteStringAsync str
```
64J0: Indeed, this is equivalent since `WriteStringAsync` implementation is:

```fsharp
/// <summary>
/// Writes an UTF-8 encoded string to the body of the HTTP response and sets the HTTP Content-Length header accordingly.
/// </summary>
/// <param name="ctx">The current http context object.</param>
/// <param name="str">The string value to be send back to the client.</param>
/// <returns>Task of Some HttpContext after writing to the body of the response.</returns>
[<Extension>]
static member WriteStringAsync(ctx: HttpContext, str: string) =
    ctx.WriteBytesAsync(Encoding.UTF8.GetBytes str)
```

Good catch.

64J0: > ...because now it constructs a new dictionary on each next-call. Will there be more next-calls than `routePorts` calls? Just to note that Dictionary is not thread-safe so not to put it outside `let routePorts`.

We might need to check the IL/ASM and run a benchmark to check if there's any noticeable performance improvement. Furthermore, if we make this change, I'd like to add new automated tests to assert things are still working in the long run.

64J0 (approved):

src/Giraffe/Core.fs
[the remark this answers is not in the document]
Thorium: Oh yes, that's irrelevant. I expected the lazy wouldn't call the actual constructor automatically, so that: 
```fsharp
open System.Collections.Concurrent
let x = Lazy<ConcurrentDictionary<int,int>>() // x.Force() would result null, but it doesn't
```
It seems to actually call the empty constructor on the first usage.

PR #741, opened 2026-07-08
Add groups for F# analyzers in Dependabot config

64J0:
## Description

Add fsharp-analyzers group to Dependabot configuration so both packages are updated together.

## How to test

N/A

## Related issues

N/A

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -15,3 +15,8 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    groups:
+      fsharp-analyzers-group:
+        patterns:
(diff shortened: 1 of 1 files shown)


nojaf (commented):
In my experience Dependabot does not update the dotnet tool and nuget packages in a single go. So this will always need a human tweak to settle. I don't know of any better way for this.

64J0:
> In my experience Dependabot does not update the dotnet tool and nuget packages in a single go. So this will always need a human tweak to settle. I don't know of any better way for this.

Indeed, this was a suggestion from Gemini to deal with this problem. I think it's worth giving it a try and reverting it later if it does not work as expected.

PR #751, opened 2026-08-17
Add a new function that writes a Markdown response

64J0:
## Description

With this PR I'm adding a new function that writes a Markdown response using the appropriate "content-type" value. I'm also adding new automated tests to cover this new function, and updating the DOCUMENTATION.

## How to test

Check CI.

## Related issues

N/A

### DIFF
--- a/src/Giraffe/Core.fs
+++ b/src/Giraffe/Core.fs
@@ -583,7 +583,7 @@ module Core =
     /// </summary>
     /// <param name="markdown">The Markdown string to be sent back to the client.</param>
     /// <returns>A Giraffe <see cref="HttpHandler" /> function which can be composed into a bigger web application.</returns>
-    let markdownString (markdown: string) : HttpHandler =
+    let markdown (markdown: string) : HttpHandler =
         let bytes = Encoding.UTF8.GetBytes markdown
 
 #if NET8_0_OR_GREATER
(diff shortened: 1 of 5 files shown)


src/Giraffe/Core.fs
```diff
@@ -577,6 +577,25 @@ module Core =
             ctx.SetContentType "text/html; charset=utf-8"
             ctx.WriteBytesAsync bytes
 
+    /// <summary>
+    /// Writes a Markdown string to the body of the HTTP response.
+    /// It also sets the HTTP header Content-Type to text/markdown and sets the Content-Length header accordingly.
+    /// </summary>
+    /// <param name="markdown">The Markdown string to be send back to the client.</param>
+    /// <returns>A Giraffe <see cref="HttpHandler" /> function which can be composed into a bigger web application.</returns>
+    let markdownString (markdown: string) : HttpHandler =
```
nojaf: Are you sure about the `String` suffix in the name here? Is it also xmlString or jsonString or yamlString?

nojaf (commented):
Not sure about the naming.

src/Giraffe/Core.fs
64J0: Good catch. Just `markdown` sounds better than `markdownString`. In fact, `markdownString` could be the parameter name instead.

64J0: Code updated

nojaf (approved):
Looks good!

